PR #17, opened 2020-06-11
Add yaml file for Google App Engine deployment

karenfrilya97:


.gcloudignore
```diff
@@ -0,0 +1,17 @@
+# This file specifies files that are *not* uploaded to Google Cloud Platform
```
ifhuang, line 1: Is this file necessary to be checked in?

karenfrilya97: I'm not too sure actually. But I thought it'd be nice to check this in so anyone can deploy to AppEngine without making their own version of this file.

.gcloudignore
```diff
@@ -0,0 +1,17 @@
...
+#
+.gcloudignore
+# If you would like to upload your .git directory, .gitignore file or files
+# from your .gitignore file, remove the corresponding line
+# below:
+.git
+.gitignore
+
+# Node.js dependencies:
+node_modules/
```
chshda, line 17: Might be better to add a new line at the end

karenfrilya97: Added! Thanks for catching this.

PR #21, opened 2020-06-11
Set up frontend folder

### LINKED ISSUE #18: Create frontend skeleton

hzxcaryn:
Closes #18

### What is in this PR:
Basically prepping the frontend folder for development
- Setting up of frontend skeleton code (using create-react-app), env for using TypeScript with React
- Setting up of gts and lint rules
- Add pre-push hook that triggers the linter


web/package.json
```diff
@@ -0,0 +1,59 @@
...
+    "build": "react-scripts build",
+    "test": "react-scripts test",
+    "eject": "react-scripts eject",
+    "lint": "gts check",
+    "clean": "gts clean",
+    "compile": "tsc",
+    "fix": "gts fix",
+    "prepare": "yarn run compile",
+    "pretest": "yarn run compile",
+    "posttest": "yarn run check"
```
karenfrilya97, line 32: I think this you might need to change the `posttest` script to `yarn run lint` instead, or change back the `check` command to `check`. Otherwise, running `yarn posttest` might run into an error. But you can check first. If there is no problem then it's okay.

hzxcaryn: Oh yea I didn't realise HAHA Good catch :D

web/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+Lints and check for formatting problems.
+
+### `yarn fix`
+
+Automatically fixes formatting and linting problems (if possible).
+
+### `yarn clean`
+
+Removes output files
```
karenfrilya97, line 31: This is very minor, but maybe add a full stop here to standardise?

web/package.json
```diff
@@ -19,7 +22,14 @@
...
     "build": "react-scripts build",
     "test": "react-scripts test",
-    "eject": "react-scripts eject"
+    "eject": "react-scripts eject",
+    "lint": "gts check",
+    "clean": "gts clean",
+    "compile": "tsc",
+    "fix": "gts fix",
+    "prepare": "yarn run compile",
+    "pretest": "yarn run compile",
```
ifhuang: Both prepare and pretest map to same command. Is it intended?

web/README.md
```diff
@@ -12,6 +18,22 @@ Open [http://localhost:3000](http://localhost:3000) to view it in the browser.
 The page will reload if you make edits.<br />
 You will also see any lint errors in the console.
 
+### `yarn lint`
```
ifhuang, line 21: Can you mention in the desc about they're based on gts?

web/package.json
hzxcaryn: Hmm they are both added by `npx gts init`. I just left it there because I thought it might be some google convention thing 🤔  haha

web/README.md
hzxcaryn: Done!

PR #20, opened 2020-06-11
Setup server development environment

karenfrilya97:
### What is in this PR:
Preparing the development environment for the backend, except for inserting boilerplate express code
- Adding nodemon for development automation
- Setting up of Google TypeScript Style (gts) to enforce style guide
- Adding README.md to give instructions to develop this application

PR #23, opened 2020-06-12
Add Travis CI

### LINKED ISSUE #22: Add CI
Add Travis CI to the repo to facilitate the development process. Ensures that all PRs passes all tests and style checks before they can be merged.

hzxcaryn:
Closes #22 

**TO BE MERGED ONLY AFTER #20 and #21**

Added basic config for travis.
CD will be set up in another PR :)


.travis.yml
```diff
@@ -0,0 +1,19 @@
...
+cache: yarn
+install:
+  - yarn
+matrix:
+  include:
+    - name: 'Backend'
+      before_install:
+        - cd server
+      script:
+        - yarn lint
```
ifhuang, line 12: should we also add test/build for backend?

karenfrilya97: yes, we should. but I don't know what to put for the test/build for backend. so I didn't add these scripts yet. I was planning to figure this out later. but if you can help us figure it out, that'll be great!

hzxcaryn: @ifhuang Yup we should, I will add it in when it's ready! I was looking at #20 `server/package.json` file and it doesn't look like the test script is up, so I left it out for now so that it doesn't break anything when this is merged :p

PR #24, opened 2020-06-12
Set up backend apis layer to handle routing of REST API request

karenfrilya97:


karenfrilya97:
> Is there a way to diffbase? I want to see this PR diff on your previous PR

For now, I changed the base branch to the branch in my previous PR. Now you can see the difference from my previous PR. But I think I need to change back the base branch to `master` before merging later.

Another way is to still use `master` as the base branch and you can view changes only from the commits after the last commit in my previous PR. But then that would be more inconvenient for you.

Unfortunately, I don't know of a better way than either of these.

karenfrilya97:
> Is there a way to diffbase? I want to see this PR diff on your previous PR

Should be good now, since I already merged my previous PR.

server/src/api/commits/index.ts
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * @fileoverview Handles routing of /commits endpoints.
+ * @author Karen Frilya Celine
+ */
+
+import * as express from 'express';
+const app: express.Application = express();
+const router: express.Router = express.Router();
+
+export const commitRouter = router;
```
ifhuang, line 26: be consistent, call it commitsRouter? same for other files

server/src/api/index.ts
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * @fileoverview Handles routing of the different API routes.
+ * @author Karen Frilya Celine
+ */
+
+import { merchantRouter } from './merchants'
+import { customerRouter } from './customers'
+import { listingRouter } from './listings'
+import { commitRouter } from './commits'
```
ifhuang, line 25: sort imports?

PR #25, opened 2020-06-12
Separate app.yaml files for client and server

hzxcaryn:
Instead of having one `app.yaml` file at the root of the project, we will create 2 separate services for the client and server side code instead.


web/app.yaml
```diff
@@ -0,0 +1,9 @@
+runtime: nodejs10
+service: default
+
+handlers:
+  - url: /
+    static_files: build/index.html
+    upload: build/index.html
+  - url: /
```
ifhuang, line 8: nit: add empty line above (following the example)

hzxcaryn: @ifhuang What example are you referring to? I think it's abit odd to put a line break because they are both part of `handlers` though!

ifhuang: https://cloud.google.com/appengine/docs/standard/python/config/appref#example

hzxcaryn: Ohh okay I'll follow that then hahaha

PR #39, opened 2020-06-16
Fix server start script for production

karenfrilya97:
Changes made:
- Changed `yarn start` command to run compiled code for production
- Added `yarn start:dev` command to run current code for development

Previously `yarn start` command ran code in development mode which caused error when deployed to the server. This PR fixes that problem.

### DIFF
--- a/server/src/index.ts
+++ b/server/src/index.ts
@@ -37,6 +37,6 @@ router.use('/merchants', merchantsRouter);
 
 app.use('/', router);
 
-const port = process.env.port || 5000;
+const port = process.env.PORT || 8080;
 app.listen(port);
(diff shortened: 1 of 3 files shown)


server/src/index.ts
```diff
@@ -37,6 +37,6 @@ router.use('/merchants', merchantsRouter);
 
 app.use('/', router);
 
-const port = process.env.port || 5000;
+const port = process.env.PORT || 8080;
```
hzxcaryn, line 40: Whyy :(( My vscode using this port now leh HAHA

karenfrilya97: okay i can change back to 5000 hahaha

PR #48, opened 2020-06-18
Add Card components

hzxcaryn:
## What's in this PR

Screenshot 2020-06-18 at 2 26 41 PM

- Added Card and ListingCard components, as well as smaller components used to build it.
- Added a temporary `/design-samples` endpoint for us to place our component designs there for testing and previewing before they are added into the app. Will be removed or kept as a dev env only endpoint in the future.

## How to test

Currently built for mobile view so it will look really stretched and ugly if you don't change it to mobile view haha.

### Test on prod

Because I deployed it haha
[https://gpay-group-buy.an.r.appspot.com/design-samples](https://gpay-group-buy.an.r.appspot.com/design-samples)

OR

### Test locally

Fetch branch then
```
cd web
yarn
yarn start
```
And then navigate to `http://localhost:3000/design-samples` and you should see the page in the screenshot above :)

### DIFF
--- a/web/src/components/common/Card.tsx
+++ b/web/src/components/common/Card.tsx
@@ -20,6 +20,8 @@ import CardImage from 'components/common/CardImage';
 import {Image} from 'interfaces';
 import styled from 'styled-components';
 
+const DEFAULT_IMG_WIDTH = '150px';
+
 interface CardContainerProps {
   imgLeft?: boolean;
 }
@@ -50,6 +52,9 @@ interface CardProps {
   imgLeft?: boolean;
 }
 
+/**
+ * A base Card component that takes in images.
+ */
 const Card: React.FC<CardProps> = ({img, imgLeft, children}) => (
   <CardContainer imgLeft={imgLeft}>
     {img !== undefined && (
--- a/web/src/components/common/Card.tsx
+++ b/web/src/components/common/Card.tsx
@@ -23,13 +23,13 @@ import styled from 'styled-components';
 const DEFAULT_IMG_WIDTH = '150px';
 
 interface CardContainerProps {
-  imgLeft?: boolean;
+  horizontal?: boolean;
 }
 
 const CardContainer = styled.div`
   display: flex;
-  flex-direction: ${({imgLeft}: CardContainerProps) =>
-    imgLeft ? 'row' : 'column'};
+  flex-direction: ${({horizontal}: CardContainerProps) =>
+    horizontal ? 'row' : 'column'};
 
   background: white;
   margin: 0.6em;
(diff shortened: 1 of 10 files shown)


web/src/index.css
```diff
@@ -24,8 +25,9 @@
   --bright-red: #EB5757;
   --red: #DB4437;
   --pale-gray: #EFEFEF;
-  --light-gray: #BDBDBD;
-  --gray: #A0A0A0;
+  --pale-gray: #BDBDBD;
```
karenfrilya97, line 28: why are there 2 different `pale-gray`s

hzxcaryn: omg thanks for noticing LOL I copy pasted and forgot to replace

karenfrilya97: HAHA okay!

web/src/components/common/Card.tsx
```diff
@@ -0,0 +1,68 @@
...
+ */
+
+import React from 'react';
+
+import CardImage from 'components/common/CardImage';
+import {Image} from 'interfaces';
+import styled from 'styled-components';
+
+interface CardContainerProps {
+  imgLeft?: boolean;
```
karenfrilya97, line 24: I think `imgLeft` sounds like whether the image is on the left or on the right. I would use variable name like `orientation: landscape` or `isLandscape` instead. I think that would be clearer.

hzxcaryn: Haha I was using `horizontal` but idk if it makes sense in this context so I changed to `imgLeft` 🤣. And I didn't wanna bring in an enum since we are only giving 2 options here hmm

karenfrilya97: Ya I also thought of `horizontal` at first but I think `landscape` makes more sense in this context. I think `imgLeft` makes sense once you see the 2 layouts, only before you see that it's a bit unclear.

web/src/components/common/StrippedCol.tsx
```diff
@@ -0,0 +1,24 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Col from 'muicss/lib/react/col';
+import styled from 'styled-components';
+
+const StrippedCol = styled(Col)`
```
ifhuang, line 20: Can you add doc/comment for this, as it's exported and used by others?
Same for other files.

web/src/components/common/Card.tsx
ifhuang: +1 for 'imgLeft' seems not very clear. When I look at the interface, I'm not sure what does imgLeft=false mean

hzxcaryn: @ifhuang So what word do you prefer? hahaha 😆

web/src/components/common/Card.tsx
```diff
@@ -0,0 +1,73 @@
...
+  flex-grow: 1;
+`;
+
+interface CardProps {
+  img?: Image;
+  horizontal?: boolean;
+}
+
+/**
+ * A base Card component that takes in images.
```
hzxcaryn, line 56:
```suggestion
 * A base Card component that can also display images.
```

PR #49, opened 2020-06-18
Fix FE app.yaml for deployment

hzxcaryn:
Previously my handlers are not inclusive enough and app routing didn't work for prod :( But I've fixed it! 😄

### DIFF
--- a/web/app.yaml
+++ b/web/app.yaml
@@ -2,9 +2,13 @@ runtime: nodejs10
 service: default
 
 handlers:
-  - url: /
+  - url: /static
+    static_dir: build/static
+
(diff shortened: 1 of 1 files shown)


web/app.yaml
```diff
@@ -2,9 +2,13 @@ runtime: nodejs10
...
 handlers:
-  - url: /
+  - url: /static
+    static_dir: build/static
+
+  - url: /(.*\.(json|ico|js|png))$
+    static_files: build/\1
+    upload: build/.*\.(json|ico|js|png)$
+
+  - url: .*
```
hzxcaryn, line 12:
```suggestion
  - url: /.*
```

PR #50, opened 2020-06-18
Add commit progress bar

hzxcaryn:
image

## What's in this PR
Creating of the commit progress bar reusable component. The `Commiters` text can be placed at the top or bottom or hidden. Screenshot shows when it is at the bottom.

### DIFF
--- a/web/src/components/common/CommitProgress.tsx
+++ b/web/src/components/common/CommitProgress.tsx
@@ -19,14 +19,9 @@ import React from 'react';
 import {Line} from 'rc-progress';
 import styled from 'styled-components';
 
-/**
- * Position of the commit text relative to the progress bar.
- */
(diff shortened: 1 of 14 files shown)


web/src/components/common/CommitProgress.tsx
```diff
@@ -0,0 +1,85 @@
...
+      {textPos && (
+        <div>
+          Commiters:{' '}
+          <b>
+            {numCommits}/{minCommits}
+          </b>
+        </div>
+      )}
+      <Line
+        percent={percent}
```
ifhuang, line 74: How does it look like when percent>100?

hzxcaryn: @ifhuang  It will just be max bar! :)

PR #51, opened 2020-06-18
Create Sign Up page for merchants

karenfrilya97:
Screenshot 2020-06-19 at 12 11 38 AM

### What is in This PR
Created UI for merchant's Sign Up page
Tested this only on laptop, will create another PR in the future to make responsive design for mobile and large monitor

On laptop:
image


web/src/components/common/FormRow.tsx
```diff
@@ -0,0 +1,63 @@
...
+    outline: none;
+  }
+`;
+
+interface FormRowProps {
+  label: string;
+  inputType: string;
+}
+
+const FormRow: React.FC<FormRowProps> = ({label, inputType}) => (
```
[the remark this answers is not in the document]
karenfrilya97: Done!

web/src/components/common/GroupBuyMerchantHeader.tsx
```diff
@@ -0,0 +1,55 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React from 'react';
+
+import styled from 'styled-components';
+
+import GPayLogo from 'assets/gpay-logo.svg';
```
hzxcaryn, line 21: I think we should consider importing SVGs as React Components so we can use them as-is instead of rendering them as images. You can do so by:
```
import { ReactComponent as GPayLogo } from 'assets/gpay-logo.svg';
```
And you could use it directly like `<GPayLogo />` without the image tag

web/src/components/common/FormRow.tsx
```diff
@@ -0,0 +1,67 @@
...
+  font-size: 14px;
+  padding: 0 15px;
+
+  border-radius: 15px;
+  border: none;
+  box-shadow: 4px 2px 4px rgba(0, 0, 0, 0.25);
+
+  &:focus {
+    outline: none;
+  }
```
hzxcaryn, line 48: Since you are stripping original `input` tag styles here (which we won't want globally anw) I think we can move this stripping to our global `index.css` file! :)

web/src/components/merchant/sign-up/SignUpCard.tsx
```diff
@@ -0,0 +1,92 @@
...
+const StyledRow = styled(Row)`
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  justify-content: center;
+
+  margin: 15px 0;
+`;
+
+const SubmitButton = styled.input`
```
hzxcaryn, line 46: Shouldnt this be a `button` instead? But consider using this [Button](https://www.muicss.com/docs/v1/react/buttons) and modifying the color and border-radius instead! This way we don't need the `cursor: pointer` and `outline: none` styles too

web/src/components/merchant/sign-up/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+import React from 'react';
+
+import SignUpCard from 'components/merchant/sign-up/SignUpCard';
+import Container from 'muicss/lib/react/container';
+import styled from 'styled-components';
+
+const PageContainer = styled(Container)`
+  max-height: 800px;
+  height: 100vh;
```
hzxcaryn, line 25:
```suggestion
  min-height: 100vh;
  height: 100%;
```
Hmm why not this? This way the sign up card will always be centralised and when content is longer than 100vh you can scroll too.

web/src/components/merchant/sign-up/SignUpCard.tsx
```diff
@@ -0,0 +1,92 @@
...
+      <FormRow label="Email" inputType="email" />
+      <FormRow label="Password" inputType="password" />
+      <FormRow label="Confirm Password" inputType="password" />
+      <FormRow label="VPA" inputType="email" />
+      <StyledRow>
+        <SubmitButton type="submit" value="Sign Up" />
+      </StyledRow>
+    </Form>
+    <div>
+      Already have an account? <a href="/merchant/sign-in">Sign in</a> now!
```
hzxcaryn, line 87: Could we use `Link` instead?
```
import { Link } from 'react-router-dom';
```
```suggestion
      Already have an account? <Link to="/merchant/sign-in">Sign in</Link> now!
```

web/src/components/merchant/sign-up/index.tsx
karenfrilya97: I think when it's centralised in the large monitor it's too low. That's why I set the max-height.

web/src/components/merchant/sign-up/SignUpCard.tsx
karenfrilya97: I thought using input type=submit is easier to submit the form data.

web/src/components/merchant/sign-up/index.tsx
karenfrilya97: Like I literally have to move my head to look at the form from the title of the tab :o

web/src/components/merchant/sign-up/SignUpCard.tsx
karenfrilya97: But okay I think Button better can put onaction also.

web/src/components/common/FormRow.tsx
karenfrilya97: okay!

web/src/components/common/GroupBuyMerchantHeader.tsx
karenfrilya97: done!

web/src/components/merchant/sign-up/index.tsx
karenfrilya97: Okay actually I think I will make the font-size and all the sizes responsive for both large monitor and mobile in a separate PR.

web/src/components/merchant/sign-up/SignUpCard.tsx
hzxcaryn: Haha technically button in a form also also do `type="submit"` but it will trigger a page refresh which is not desired! So yep just handle form submissions with the `onClick` :)

web/src/components/common/FormRow.tsx
```diff
@@ -0,0 +1,63 @@
...
+}
+
+/**
+ * This is a row in a form, consisting of a label and an input field shown
+ * side by side.
+ */
+const FormRow: React.FC<FormRowProps> = ({label, inputType}) => (
+  <StyledRow>
+    <Label>{label}</Label>
+    <Input type={inputType}></Input>
```
hzxcaryn, line 59: We miggght wanna add a `name` field for the input tag!

web/src/components/common/GroupBuyMerchantHeader.tsx
```diff
@@ -0,0 +1,55 @@
...
+const Logo = styled(GPayLogo)`
+  width: 80px;
+`;
+
+const Header = styled.h1`
+  font-size: 32px;
+  font-weight: normal;
+  line-height: 48px;
+  margin: 0;
+`;
```
hzxcaryn, line 38: Btw I think this is missing!
```suggestion
const Header = styled.h1`
  font-size: 32px;
  font-weight: normal;
  line-height: 48px;
  margin: 0;
  text-align: center;
`;
```

hzxcaryn: Actually I think it should not be a `1` tag because `h1`s describes the main content of the page, for this the content should be "Sign Up"/"Merchant Sign Up" Instead.
But not sure if it's important, can always change next time also HAHA

web/src/components/common/FormRow.tsx
karenfrilya97: Hahahah I was just adding that!

karenfrilya97: I'll add to the next PR okay!

PR #52, opened 2020-06-19
Add functions to check that Merchant Sign Up form data are valid

karenfrilya97:
Added functionality to
- Check if any field is empty
- Check if password and confirm password match
- Display the corresponding error message below the field

These checks are performed when a user is typing and when they click on the 'SIGN UP' button.

image


web/src/components/merchant/sign-up/SignUpCard.tsx
```diff
@@ -56,31 +63,142 @@ const StyledButton = styled(Button)`
...
+    }
+
+    if (password !== confirmPassword) {
+      this.setState({
+        errorMessage: 'Passwords do not match.',
+      });
+      return;
+    }
+
+    // Add a new user account to Firebase.
```
karenfrilya97, line 127: I think that ideally, adding user account to Firebase shouldn't be in this file. But I don't know where to save it. Should I create src/services/sign-up-firebase.tsx file?

karenfrilya97:
@hzxcaryn said I should use react hooks instead

web/src/components/merchant/sign-up/SignUpCard.tsx
ifhuang, line 127: 1. Why do we need this?
2. Can we just call server after submission, and move this logic to server side?

karenfrilya97: 1. We are using Firebase authentication, so next time when user signs in, Firebase can do the authentication for us.
I will add a comment in the code explaining this.

2. We can but since we are sending password to server, we then have to write a code to hash the password ourselves. Here, Firebase client library does that for us so we don't have to do a lot of work. We can just send a token that they give us to the server, and then the server side verify that with the Firebase.

karenfrilya97: I am moving this part to a new PR because I realised I am doing 2 things in this PR.

web/src/components/merchant/sign-up/SignUpCard.tsx
```diff
@@ -56,31 +65,123 @@ const StyledButton = styled(Button)`
...
+  const [password, setPassword] = useState('');
+  const [passwordError, setPasswordError] = useState('');
+
+  const [confirmPassword, setConfirmPassword] = useState('');
+  const [confirmPasswordError, setConfirmPasswordError] = useState('');
+
+  const [vpa, setVpa] = useState('');
+  const [vpaError, setVpaError] = useState('');
+
+  const handleNameChange = (event: React.ChangeEvent<HTMLInputElement>) => {
```
ifhuang, line 91: prefer to move this above to put together with name variables. probably a separated function?

web/src/components/merchant/sign-up/SignUpCard.tsx
```diff
@@ -56,31 +65,123 @@ const StyledButton = styled(Button)`
...
+      : setConfirmPasswordError(PASSWORDS_DO_NOT_MATCH);
+    setConfirmPassword(event.target.value);
+  };
+
+  const handleVpaChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    event.target.value ? setVpaError('') : setVpaError(VPA_EMPTY);
+    setVpa(event.target.value);
+  };
+
+  const handleSubmit = () => {
```
ifhuang, line 130: can you enable the submit button only if there is no error? then no need to have this check?

karenfrilya97: The submit button triggers some checks also, so I think that might be a bit weird.

web/src/components/merchant/sign-up/SignUpCard.tsx
karenfrilya97: What do you mean by a separated function? The reason why I put this here is because it's a function and I want the above to show only the states, all together.

web/src/constants/sign-up-errors.ts
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export const NAME_EMPTY = 'Name cannot be empty.';
+export const EMAIL_EMPTY = 'Email cannot be empty.';
+export const PASSWORD_EMPTY = 'Password cannot be empty.';
+export const PASSWORDS_DO_NOT_MATCH = 'Passwords do no match.';
```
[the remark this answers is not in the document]
karenfrilya97: Oh thanks for catching that!

web/src/components/merchant/sign-up/SignUpCard.tsx
karenfrilya97: For example, when user's email is already in use by existing account, then the error is only thrown after the user clicks the submit button

karenfrilya97: Okay, actually after thinking about it, it makes sense to enable the submit button only if there is no detectable error at that point. But it requires quite a bit of time to implement. I just feel that this is not a necessary feature and can be implemented later. I think that I should prioritise on other more important features first.

web/src/components/common/FormRow.tsx
```diff
@@ -44,19 +58,49 @@ const Input = styled.input`
...
+`;
+
 interface FormRowProps {
   label: string;
   inputType: string;
+  onChange: React.EventHandler<React.FormEvent<HTMLInputElement>>;
+  error: string;
 }
 
+const toCamelCase = (phrase: string) =>
```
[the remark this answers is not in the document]
karenfrilya97: okay!

web/src/components/merchant/sign-up/SignUpCard.tsx
```diff
@@ -14,7 +14,15 @@
  * limitations under the License.
  */
 
-import React from 'react';
+import {
+  NAME_EMPTY,
+  EMAIL_EMPTY,
+  PASSWORD_EMPTY,
+  PASSWORDS_DO_NOT_MATCH,
+  VPA_EMPTY,
```
xlai20, line 22: Are these all the entries we need to onboard a merchant? How about phone number?
What is the identifier for this merchant? Which field is allowed to be null and why?

web/src/components/merchant/sign-up/SignUpCard.tsx
```diff
@@ -56,31 +65,123 @@ const StyledButton = styled(Button)`
...
-const SignUpCard: React.FC = () => (
-  <CardContainer>
-    <GroupBuyMerchantHeader />
-    <Form>
-      <FormRow label="Name" inputType="text" />
-      <FormRow label="Email" inputType="email" />
-      <FormRow label="Password" inputType="password" />
-      <FormRow label="Confirm Password" inputType="password" />
-      <FormRow label="VPA" inputType="email" />
+const SignUpCard: React.FC = () => {
```
xlai20, line 75: Could you organize all these data accessors/setters/functions of different fields in a custom hook? https://reactjs.org/docs/hooks-custom.html
And then package them into a single UI component as SignUpCard.

web/src/components/merchant/sign-up/SignUpCard.tsx
karenfrilya97: Yes, those are all the entries needed. Actually, I'm not sure if we still need to collect the user's VPA since we are no longer using Spot Payments API.
I don't see why we would need a phone number. But did I miss something?
The identifier would be the email.
None of the fields are allowed to be null because everything is needed.

web/src/components/common/FormRow.tsx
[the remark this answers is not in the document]
ifhuang: we'd better to move this to a separated util file (not UI related). then it's cleaner and easier to write test.

web/src/components/merchant/sign-up/SignUpCard.tsx
ifhuang: we'd better to group by logic instead of state then function.

web/src/components/merchant/sign-up/SignUpCard.tsx
ifhuang: ok. I probably left same comment in a different PR

web/src/components/merchant/sign-up/SignUpCard.tsx
xlai20: This is a custom hook, which decouples UI and logic. Since there are too many fields here, I only pick name and email as examples.
```javascript
function useSignupCard() {                                                         
  const [state, setState] = useState({                                             
    name: '',                                                                      
    nameError: '',                                                                 
    email: '',                                                                     
    emailError: ''                                                                 
  });                                                                              
                                                                                   
  const handleEmailChange = (event: React.ChangeEvent<HTMLInputElement>) => {   
    setState(Object.assign(state, {                                                
      email: event.target.value,                                                   
      emailError: event.target.value ? '': EMAIL_EMPTY,                            
    }));                                                                           
  }                                                                                
                                                                                   
  return {                                                                         
    state,                                                                         
    setSate,                                                                       
    handleEmailChange,                                                             
  }                                                                                
}};
```

web/src/components/common/FormRow.tsx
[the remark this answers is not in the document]
karenfrilya97: okay!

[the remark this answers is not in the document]
karenfrilya97: I decided to use a library instead so we have less things to test.

web/src/components/merchant/sign-up/SignUpCard.tsx
karenfrilya97: I managed to add this feature! Because I am now using a library so adding this feature becomes very easy.

web/src/components/merchant/sign-up/SignUpCard.tsx
karenfrilya97: I managed to use react-hook-form library now so most of the logic is done by the library.

web/src/components/common/FormRow.tsx
```diff
@@ -44,19 +60,57 @@ const Input = styled.input`
...
+      | string
+      | ((
+          props: any
+        ) => React.ReactElement<
+          any,
+          string | any | (new (props: any) => React.Component<any, any, any>)
+        > | null)
+      | (new (props: any) => React.Component<any, any, any>)
+    >
+  | undefined;
```
hzxcaryn, line 86: Oh I took a look and you should actually:
```
import {FieldError, NestDataObject} from 'react-hook-form';
```
```suggestion
type ReactHookFormErrorMessage = string | NestDataObject<any, FieldError> | undefined;
```

web/src/components/merchant/sign-up/SignUpForm.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+const SignUpForm = () => {
+  const {disabled, errors, onSubmit, validations} = useSignUpForm();
+  return (
+    <StyledForm>
+      <FormRow
+        label="Name"
+        inputType="text"
+        forwardedRef={validations.name}
+        error={errors?.name?.message}
```
hzxcaryn, line 53: Why does `errors` need optional chaining here? If I'm not wrong I think only the fields are possibly undefined? Same for the rest of the form rows!
```suggestion
        error={errors.name?.message}
```

web/src/components/common/FormRow.tsx
karenfrilya97: ohh thanks for looking into this!

web/src/components/merchant/sign-up/SignUpForm.tsx
karenfrilya97: ohh. I thought that errors might be undefined also. But I think you're right. Only the fields are possibly undefined.

web/src/components/merchant/sign-up/hooks/useSignUpForm.tsx
```diff
@@ -0,0 +1,85 @@
...
+    vpa: register({
+      required: VPA_EMPTY,
+      pattern: {
+        value: /^[A-Z0-9]+@[A-Z0-9]+/i,
+        message: VPA_INVALID,
+      },
+    }),
+  };
+  const disabled = !formState.isValid;
+  const onSubmit = handleSubmit((values: SignUpData) => console.log(values));
```
hzxcaryn, line 75: Oh is this console.log intentional or leftover from debugging? HAHA

karenfrilya97: Hmm it's like a dummy submit function. Will replace with a proper one with the next PR.

PR #54, opened 2020-06-22
Add Firebase authentication for Merchant Sign Up

karenfrilya97:
We decided to use Firebase to handle authentication since it has a lot of authentication capabilities.
This PR adds Firebase code to handle authentication when merchant signs up.
This also stores merchant's account to Firebase so merchant can sign in with then same account in the future.

Some additional feature given by Firebase Authentication is checking email format, password strength and if existing user has the same email.


web/app.yaml
```diff
@@ -12,3 +12,8 @@ handlers:
   - url: /.*
     static_files: build/index.html
     upload: build/index.html
+
+env_variables:
+  # Replace with API key and authentication domain before deployment
```
ifhuang, line 17: Can we create a readme file as deployment playbook? (future PR)

web/src/firebase-auth.ts
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+import firebase from 'firebase';
+
+const config = {
+  apiKey: process.env.REACT_APP_API_KEY,
+  authDomain: process.env.REACT_APP_AUTH_DOMAIN,
+};
+
+if (!firebase.apps.length) {
```
ifhuang, line 24: what does this check mean?

web/src/constants/sign-up-errors.ts
```diff
@@ -14,8 +14,28 @@
...
-export const VPA_EMPTY = 'VPA cannot be empty.';
+const NAME_EMPTY = 'Name cannot be empty.';
+const EMAIL_EMPTY = 'Email cannot be empty.';
+const PASSWORD_EMPTY = 'Password cannot be empty.';
+const VPA_EMPTY = 'VPA cannot be empty.';
+
+const PASSWORDS_DO_NOT_MATCH = 'Passwords do not match.';
+
+const INVALID_EMAIL = 'Please input a valid email address.';
+const WEAK_PASSWORD = 'Password should be at least 6 characters.';
```
ifhuang, line 25: Why this error? Is there a error msg sent from firebase?

web/src/components/merchant/sign-up/SignUpCard.tsx
```diff
@@ -118,24 +118,50 @@ const SignUpCard: React.FC = () => {
...
     setConfirmPassword(event.target.value);
   };
 
   const handleVpaChange = (event: React.ChangeEvent<HTMLInputElement>) => {
-    event.target.value ? setVpaError('') : setVpaError(VPA_EMPTY);
+    event.target.value ? setVpaError('') : setVpaError(Errors.VPA_EMPTY);
     setVpa(event.target.value);
   };
 
   const handleSubmit = () => {
```
ifhuang, line 130: can you disable the submit button if there is some error

web/app.yaml
karenfrilya97: Okay! Created issue #69.

web/src/constants/sign-up-errors.ts
karenfrilya97: Yeah firebase has these errors.

web/src/components/merchant/sign-up/SignUpCard.tsx
```diff
@@ -118,24 +118,50 @@ const SignUpCard: React.FC = () => {
...
+    if (!name) setNameError(Errors.NAME_EMPTY);
+    if (!email) setEmailError(Errors.EMAIL_EMPTY);
+    if (!password) setPasswordError(Errors.PASSWORD_EMPTY);
     if (password !== confirmPassword)
-      setConfirmPasswordError(PASSWORDS_DO_NOT_MATCH);
-    if (!vpa) setVpaError(VPA_EMPTY);
+      setConfirmPasswordError(Errors.PASSWORDS_DO_NOT_MATCH);
+    if (!vpa) setVpaError(Errors.VPA_EMPTY);
+
+    if (!name || !email || !password || password !== confirmPassword || !vpa) {
```
xlai20, line 140: Is it possible to avoid doing a second time check here? If there is a growing list of validation checks (e.g., 20 items), this is a long line of repetition.

web/src/components/merchant/sign-up/SignUpCard.tsx
```diff
@@ -118,24 +118,50 @@ const SignUpCard: React.FC = () => {
...
+
+  /**
+   * This adds a new user account to Firebase so that user can be signed in with
+   * Firebase Authentication in the future.
+   */
+  const handleFirebaseSignUp = async () => {
+    try {
+      await firebaseAuth.createUserWithEmailAndPassword(email, password);
+    } catch (error) {
+      if (error.code === 'auth/invalid-email') {
```
xlai20, line 155: Can change to switch statement?

"Generally speaking, if-else is best used when there are two discrete values or a few different ranges of values for which to test. When there are more than two discrete values for which to test, the switch statement is the most optimal choice" - [source](https://www.oreilly.com/library/view/high-performance-javascript/9781449382308/ch04.html#if-else_versus_switch)

web/src/firebase-auth.ts
karenfrilya97: It's to prevent re-initializing the firebase app when it's already created.
Actually, maybe I should change to 
```suggestion
if (firebase.apps.length > 0) {
```

web/src/components/merchant/sign-up/SignUpCard.tsx
karenfrilya97: Okay!

web/src/components/merchant/sign-up/SignUpCard.tsx
karenfrilya97: Changed to using library so we don't have to do this.

web/src/components/merchant/sign-up/SignUpCard.tsx
karenfrilya97: Okay! Yep, switch-case is better in this case.

PR #57, opened 2020-06-22
Add stop handler on termination signals

hzxcaryn:
Force processes to be killed because previously we kept getting errors like:
```
Error: listen EADDRINUSE: address already in use :::5000
...
[nodemon] app crashed - waiting for file changes before starting...
```

### DIFF
--- a/server/src/index.ts
+++ b/server/src/index.ts
@@ -68,7 +68,7 @@ app.use(
 const port = process.env.PORT || 5000;
 app.listen(port, () => console.log(`Listening on port ${port}`));
 
-const stopHandler = () => process.exit(1);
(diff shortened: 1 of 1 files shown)


server/src/index.ts
```diff
@@ -66,5 +66,10 @@ app.use(
 );
 
 const port = process.env.PORT || 5000;
-app.listen(port);
-console.log(`Listening on port ${port}`);
+app.listen(port, () => console.log(`Listening on port ${port}`));
+
+const stopHandler = () => process.exit(1); // eslint-disable-line no-process-exit
```
ifhuang, line 71: can you also add a log for this?

PR #59, opened 2020-06-23
Add lazy loading for images

hzxcaryn:
## Changes
- Added a `LazyWrapper` that only renders children components when the component is in view.
- Wrapped `ListingCard` images with `LazyWrapper` to lazy load images for performance

### DIFF
--- a/web/src/components/common/LazyWrapper.tsx
+++ b/web/src/components/common/LazyWrapper.tsx
@@ -18,10 +18,17 @@ import React from 'react';
 
 import {useInView} from 'react-hook-inview';
 
-const LazyWrapper: React.FC = ({children}) => {
+interface LazyWrapperProps {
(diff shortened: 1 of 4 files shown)


web/src/components/common/LazyWrapper.tsx
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+import React from 'react';
+
+import {useInView} from 'react-hook-inview';
+
+const LazyWrapper: React.FC = ({children}) => {
+  const [ref, isInView] = useInView({
+    unobserveOnEnter: true,
+    rootMargin: '30px',
```
ifhuang, line 24: What is this? Is it used to prefetch before children is visible in the window?
Can we add a function param for this with default value 30px?

hzxcaryn: @ifhuang Yep that's right, the `child` will be fetched 30px before it is in view! Ok sure I can make it a prop :)

PR #61, opened 2020-06-23
Create add merchant API

karenfrilya97:


server/src/services/merchants.ts
```diff
@@ -14,10 +14,10 @@
  * limitations under the License.
  */
 
+import {MerchantPayload} from '../interfaces';
 import {merchantStorage} from '../storage';
 
-const getAllMerchants = async () => await merchantStorage.getAllMerchants();
+const addMerchant = async (merchant: MerchantPayload) =>
+  await merchantStorage.addMerchant(merchant);
```
hzxcaryn, line 21:
```suggestion
  merchantStorage.addMerchant(merchant);
```
Actually you won't need await here!
Read: https://eslint.org/docs/rules/no-return-await

server/src/storage/merchants.ts
```diff
@@ -17,36 +17,14 @@
...
-  return await datastore.get(key);
-}
-
-async function addMerchant(merchant: object) {
-  return await datastore.insert({
-    key: datastore.key([MERCHANT_KIND, Guid.raw()]),
-    data: merchant,
-  });
+async function addMerchant(merchant: MerchantPayload) {
+  await add(MERCHANT_KIND, merchant);
```
hzxcaryn, line 27: Hmm do we need any response from this?

server/src/services/merchants.ts
karenfrilya97: Oh! Because the arrow notation has implicit return right?

server/src/storage/merchants.ts
karenfrilya97: I don't think so, you just need to `await`. If there is any error it would throw an error, which will eventually be caught at the api/merchants.ts. Otherwise, then it is successfully added.

server/src/services/merchants.ts
hzxcaryn: Yep arrow function w/ the braces is an implicit return so over here what's going on now is equivalent to a `return await`!

server/src/storage/datastore.ts
```diff
@@ -33,4 +33,15 @@ const getWithId = async (kind: string, id: string) => {
   };
 };
 
-export {getWithId};
+/**
+ * A Datastore wrapper that inserts a particular entity with the specified Kind.
+ * @param kind The Kind of the Entity
+ * @param data The data of the Entity to be added
+ */
+const add = async (kind: string, data: object) => {
```
ifhuang, line 41: How will you prevent same data to be added multiple times?
Let's say merchant clicks 'sign up' button twice with the same form data

server/src/interfaces.ts
```diff
@@ -31,3 +31,22 @@ export interface CustomerPayload {
...
+
+/**
+ * MerchantPayload Interface that contains the fields of the payload that
+ * would be sent to create a Merchant Entity.
+ */
+export interface MerchantPayload {
+  name: string;
+  email: string;
+  vpa: string;
+  uid: string;
```
ifhuang, line 43: What is this?

server/src/api/merchants.ts
```diff
@@ -18,10 +18,24 @@
...
+const merchantRouter: Router = Router();
 
-export const merchantRouter: express.Router = router;
+merchantRouter.post(
+  '/',
+  async (req: Request, res: Response, next: NextFunction) => {
+    try {
+      const merchant: MerchantPayload = req.body;
+      await merchantService.addMerchant(merchant);
+      res.sendStatus(200);
```
ifhuang, line 34: No need to send back merchant id?

karenfrilya97: No need. The frontend has `idToken` from Firebase Authentication. They should attach it as Authorization header to the backend. This will the one used to identify the merchant.

server/src/interfaces.ts
karenfrilya97: This is the unique ID of the user from Firebase Authentication. Actually, now that I think of it, maybe I should rename it to `firebaseUid`.

server/src/storage/datastore.ts
karenfrilya97: I haven't added checks for that yet. Was planning to do it in another PR. 

Anyway, my plan is to make sure that the firebase `uid` is unique in the table. Users with different `uid` will have different `email` as ensured by Firebase Authentication. I know that for SQL databases we can put this restriction on the database. I'm hoping there is a similar way to do this for Datastore but I need to find out more on that. But even with that, if we want to be defensive we can still add a code to check.

server/src/storage/merchants.ts
karenfrilya97: Okay based on discussion on Chat, added a response for this to return the inserted id.

server/src/api/merchants.ts
karenfrilya97: Okay, based on discussion on Chat, will send id back to merchant.

server/src/interfaces.ts
karenfrilya97, line 43:
```suggestion
  firebaseUid: string;
```

server/src/storage/datastore.ts
```diff
@@ -33,4 +34,27 @@ const getWithId = async (kind: string, id: string) => {
...
+  const entity = {key, data};
+  const [res] = await datastore.insert(entity);
+  return getIdFromCommitResponse(res);
+};
+
+/**
+ * Unpacks id of inserted object from CommitResponse object received from Datastore.
+ */
+const getIdFromCommitResponse = (res: google.datastore.v1.ICommitResponse) => {
+  const mutationResult = res.mutationResults && res.mutationResults[0];
```
ifhuang, line 53: Can you use optional chaining? https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining

server/src/storage/datastore.ts
```diff
@@ -33,4 +34,27 @@ const getWithId = async (kind: string, id: string) => {
...
+  const key = datastore.key(kind);
+  const entity = {key, data};
+  const [res] = await datastore.insert(entity);
+  return getIdFromCommitResponse(res);
+};
+
+/**
+ * Unpacks id of inserted object from CommitResponse object received from Datastore.
+ */
+const getIdFromCommitResponse = (res: google.datastore.v1.ICommitResponse) => {
```
ifhuang, line 52: Better call it getFirstId...

server/src/storage/datastore.ts
```diff
@@ -33,4 +34,27 @@ const getWithId = async (kind: string, id: string) => {
...
+/**
+ * A Datastore wrapper that inserts a particular entity with the specified Kind.
+ * @param kind The Kind of the Entity
+ * @param data The data of the Entity to be added
+ */
+const add = async (kind: string, data: object) => {
+  const key = datastore.key(kind);
+  const entity = {key, data};
+  const [res] = await datastore.insert(entity);
+  return getIdFromCommitResponse(res);
```
ifhuang, line 46: This makes your add function nullable. Instead of that can you assert id != null or throw some exception of it's null

server/src/storage/datastore.ts
karenfrilya97: Oh right, because CommitResponse can modify multiple objects.
Actually, I decided to just change it to `getIdsFromCommitResponse` instead to make it more general.

server/src/storage/datastore.ts
karenfrilya97: Okay!

server/src/storage/datastore.ts
karenfrilya97: Okay!

server/src/storage/datastore.ts
```diff
@@ -55,4 +56,37 @@ const getAllWithId = async (kind: string) => {
   return resWithId;
 };
 
-export {getWithId, getAllWithId};
+/**
+ * Unpacks id of modified object from MutationResult object.
+ */
+const getIdFromMutationResult = (
+  mutationResult: google.datastore.v1.IMutationResult
+): number | Long | string | null | undefined =>
```
ifhuang, line 64: why this is return type? same type of id in mutationResult?

server/src/storage/datastore.ts
```diff
@@ -55,4 +56,37 @@ const getAllWithId = async (kind: string) => {
...
+const add = async (kind: string, data: object): Promise<number> => {
+  const key = datastore.key(kind);
+  const entity = {key, data};
+  const [res] = await datastore.insert(entity);
+  const ids = getIdsFromCommitResponse(res);
+  const id = ids?.[0];
+  if (id === null || id === undefined) {
+    throw new Error(`Failed to add ${kind}.`);
+  }
+  return Number(id);
```
ifhuang, line 89: What happen if id is not a number?

karenfrilya97: It should be a number because when we call `datastore.insert(entity)` where the `key` only specifies the Kind, then the id is auto-generated by Datastore, which should be a number.

server/src/storage/datastore.ts
karenfrilya97: Yep, the Datastore id has return type `number | Long | string | null`. I refactored this out into a `ResponseId` type.

PR #63, opened 2020-06-23
Add numCommits to get customer response

hzxcaryn:
## Changes
- Added numCommits to customer getter response. Num commits is retrieved by querying `Commits` and filtering by customerId and listingStatus

## How to test
Fetch my branch
```
cd server
yarn
yarn start:dev
```
Make a request to `http://localhost:5000/customers/5634161670881280` (our dear customer) and there will be a new `numCommits` field added. Our dear customer here should have 1 commit.

### DIFF
--- a/server/src/api/customers.ts
+++ b/server/src/api/customers.ts
@@ -27,10 +27,11 @@ const customerRouter = Router();
 customerRouter.get(
   '/:customerId',
   async (req: Request, res: Response, next: NextFunction) => {
-    const {customerId} = req.params;
+    const {customerIdStr} = req.params;
+    const customerId = Number(customerIdStr);
 
     try {
-      const customer = await customerService.getCustomer(Number(customerId));
+      const customer = await customerService.getCustomer(customerId);
       res.send(customer);
     } catch (error) {
       return next(error);
--- a/server/src/api/customers.ts
+++ b/server/src/api/customers.ts
@@ -27,7 +27,7 @@ const customerRouter = Router();
 customerRouter.get(
   '/:customerId',
   async (req: Request, res: Response, next: NextFunction) => {
-    const {customerIdStr} = req.params;
+    const {customerId: customerIdStr} = req.params;
     const customerId = Number(customerIdStr);
 
     try {
--- a/server/src/api/customers.ts
+++ b/server/src/api/customers.ts
@@ -30,7 +30,7 @@ customerRouter.get(
     const {customerId} = req.params;
 
     try {
-      const customer = await customerService.getCustomer(customerId);
+      const customer = await customerService.getCustomer(Number(customerId));
       res.send(customer);
     } catch (error) {
       return next(error);
--- a/server/src/interfaces.ts
+++ b/server/src/interfaces.ts
@@ -44,9 +44,9 @@ export type CommitStatus =
   | 'unsuccessful';
 
 /**
-* ListingPayload Interface that contains the fields of the payload that
-* would be sent to create a Listing Entity.
-*/
+ * ListingPayload Interface that contains the fields of the payload that
+ * would be sent to create a Listing Entity.
+ */
 export interface CommitPayload {
   customerId: number;
   listingId: number;
--- a/server/src/interfaces.ts
+++ b/server/src/interfaces.ts
@@ -30,7 +30,7 @@ export interface CustomerPayload {
  */
 export interface CustomerResponse extends CustomerPayload {
   id: number;
-  numCommits: number;
+  numOngoingCommits: number;
 }
 
 /**
--- a/server/src/interfaces.ts
+++ b/server/src/interfaces.ts
@@ -44,7 +44,7 @@ export type CommitStatus =
   | 'unsuccessful';
 
 /**
- * ListingPayload Interface that contains the fields of the payload that
+ * CommitPayload Interface that contains the fields of the payload that
  * would be sent to create a Listing Entity.
  */
 export interface CommitPayload {
@@ -55,7 +55,7 @@ export interface CommitPayload {
 }
 
 /**
- * ListingResponse Interface that contains the fields of the Response that
+ * CommitResponse Interface that contains the fields of the Response that
  * client side would receive.
  */
 export interface CommitResponse extends CommitPayload {
--- a/server/src/interfaces.ts
+++ b/server/src/interfaces.ts
@@ -15,11 +15,6 @@
  */
 
 /**
- * ResponseId type is the union.type of datastore's built-in id.
- */
-export type ResponseId = number | Long | string | null;
-
-/**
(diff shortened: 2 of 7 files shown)


server/src/interfaces.ts
```diff
@@ -30,4 +30,34 @@ export interface CustomerPayload {
...
+ */
+export type CommitStatus =
+  | 'ongoing'
+  | 'successful'
+  | 'paid'
+  | 'completed'
+  | 'unsuccessful';
+
+/**
+ * ListingPayload Interface that contains the fields of the payload that
```
karenfrilya97, line 47:
```suggestion
 * CommitPayload Interface that contains the fields of the payload that
```
You forgot to change this!

hzxcaryn: oopsies!

server/src/storage/interfaces.ts
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Filter Interface that contains the fields of the payload that
+ * would be sent to create a Customer Entity.
```
ifhuang, line 19: Filter sounds generic enough, not only for Customer Entity?

server/src/storage/datastore.ts
```diff
@@ -16,21 +16,47 @@
...
-export {getWithId};
+/**
+ * A Datastore wrapper that gets all entities of a specified Kind.
+ * @param kind The Kind that is being queried
+ * @param filters Any filters that will be applied to the query
+ */
+const getAllWithId = async (kind: string, filters?: Filter[]) => {
+  let query = datastore.createQuery(kind);
+  if (filters) {
+    filters.forEach(filter => {
```
ifhuang, line 46: can you use "filters?.", then no need if check?

server/src/storage/customers.ts
```diff
@@ -14,11 +14,27 @@
...
-import {getWithId} from './datastore';
+import {CUSTOMER_KIND, COMMIT_KIND} from '../constants/kinds';
+import {CustomerResponse, CommitResponse} from '../interfaces';
+import {getWithId, getAllWithId} from './datastore';
 
-const getCustomer = async (customerId: string): Promise<CustomerResponse> =>
-  getWithId(CUSTOMER_KIND, customerId);
+const getCustomer = async (customerId: number): Promise<CustomerResponse> => {
+  const customer = await getWithId(CUSTOMER_KIND, customerId);
+  const commits: CommitResponse[] = await getAllWithId(COMMIT_KIND, [
```
ifhuang, line 23: Will commits=null if no commits yet?

server/src/interfaces.ts
```diff
@@ -30,4 +30,34 @@ export interface CustomerPayload {
...
+  | 'unsuccessful';
+
+/**
+ * CommitPayload Interface that contains the fields of the payload that
+ * would be sent to create a Listing Entity.
+ */
+export interface CommitPayload {
+  customerId: number;
+  listingId: number;
+  createdAt: string;
```
ifhuang, line 53: date type?

server/src/storage/interfaces.ts
hzxcaryn: Omg i forgot to edit the docstring 😰 😅

server/src/storage/customers.ts
hzxcaryn: Nope will just get empty array!

server/src/interfaces.ts
hzxcaryn: Oh yea, will change for this PR too.

server/src/storage/customers.ts
```diff
@@ -14,11 +14,27 @@
...
+    },
+    {
+      property: 'commitStatus',
+      value: 'ongoing',
+    },
+  ]);
+
+  return {
+    ...customer,
+    numCommits: commits.length,
```
ifhuang, line 36: Is there a count api that we can call on the datastore?

server/src/storage/datastore.ts
```diff
@@ -38,18 +40,23 @@ const extractAndAppendId = (res: Entity) => {
...
 /**
  * A Datastore wrapper that gets all entities of a specified Kind.
  * @param kind The Kind that is being queried
+ * @param filters Any filters that will be applied to the query
  */
-const getAllWithId = async (kind: string) => {
-  const query = datastore.createQuery(kind);
+const getAllWithId = async (kind: string, filters?: Filter[]) => {
+  let query = datastore.createQuery(kind);
+  filters?.forEach(filter => {
```
ifhuang, line 56: remove {}

server/src/storage/datastore.ts
```diff
@@ -38,18 +40,23 @@ const extractAndAppendId = (res: Entity) => {
...
 };
 
 /**
  * A Datastore wrapper that gets all entities of a specified Kind.
  * @param kind The Kind that is being queried
+ * @param filters Any filters that will be applied to the query
  */
-const getAllWithId = async (kind: string) => {
-  const query = datastore.createQuery(kind);
+const getAllWithId = async (kind: string, filters?: Filter[]) => {
```
karenfrilya97, line 54: I think we should rename this to `getAll`. `getWithId` means that you want to get all the objects having the specified id right? So this one should be `getAll` or `getAllWithFilter`. But since the filter is optional then just `getAll` is better.

server/src/storage/datastore.ts
hzxcaryn: @ifhuang Oh hmmm I don't think I should do that because I'm just modifying query variable and not returning it!

server/src/storage/datastore.ts
hzxcaryn: Oh okay I meant it like `WithId` meaning the thing you get comes with the `id`. Okay but since there's some confusion I will just rename both of them then HAHA

server/src/storage/customers.ts
chshda, line 36: Looks like this is ongoing commit only, maybe rename to make that more explicit

hzxcaryn: I can't find any for node.js!

server/src/api/customers.ts
```diff
@@ -30,7 +30,7 @@ customerRouter.get(
     const {customerId} = req.params;
 
     try {
-      const customer = await customerService.getCustomer(customerId);
+      const customer = await customerService.getCustomer(Number(customerId));
```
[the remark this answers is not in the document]
hzxcaryn: I think that's fair

PR #64, opened 2020-06-23
Add merchant authentication using Firebase

karenfrilya97:


server/tsconfig.json
```diff
@@ -12,7 +12,10 @@
     "outDir": "build",
     "skipLibCheck": true,
     "strict": true,
-    "baseUrl": "./src/"
+    "baseUrl": "./src/",
+    "typeRoots": [
+      "./custom_typings"
```
hzxcaryn, line 17: Should we be consistent and use `-` case for file and folder names? :p

server/src/middleware/merchant-auth.ts
```diff
@@ -0,0 +1,39 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {Request, Response, NextFunction} from 'express';
+import admin from 'firebase-admin';
+
+admin.initializeApp();
```
hzxcaryn, line 20: Should we do something similar to the client side, creating a `src/firebase.ts` which initializes the app and exports the `const auth = admin.auth();`?

server/custom_typings/express/index.d.ts
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+declare namespace Express {
+  interface Request {
+    decoded: import('firebase-admin').auth.DecodedIdToken;
```
ifhuang, line 19: decoded -> firebaseDecodedIdToken?

server/src/middleware/merchant-auth.ts
```diff
@@ -0,0 +1,39 @@
...
+
+const merchantAuth = async (
+  req: Request,
+  res: Response,
+  next: NextFunction
+) => {
+  try {
+    if (!req.headers.authorization) throw new Error('Unauthorized');
+
+    const [_, idToken] = req.headers.authorization.split(' ');
```
ifhuang, line 30: idToken -> firebaseIdToken?

server/tsconfig.json
karenfrilya97: I was following the node_modules cos this one is the same level as that. But I guess we can just standardise all the files that we make.

server/src/middleware/merchant-auth.ts
karenfrilya97: Yep, I think I've already done that  https://github.com/googleinterns/gpay-group-buy/pull/54/commits/2f1b38df7a3993c6eb523b4b94badb447d99ecd2

server/tsconfig.json
karenfrilya97: Also, I realise I can put this folder inside src then it'll be better.

server/custom_typings/express/index.d.ts
karenfrilya97: I think the convention is `req.decoded` when you put something you decoded from a token.

server/src/middleware/merchant-auth.ts
karenfrilya97: Okay!

server/src/middleware/merchant-auth.ts
hzxcaryn: @karenfrilya97 Isn't that on the client side? I actually meant like doing the same for server side too

PR #70, opened 2020-06-26
Add individual listing getter

hzxcaryn:
Part of #16 
## Changes
Added endpoint for getting individual listing

## How to test
```
cd server
yarn 
yarn start:dev
```
Make a request to `http://localhost:5000/listings/5081054809423872`. 
This is the output I got:
```
{
    "merchantId": 5644004762845184,
    "imgUrl": "https://images.unsplash.com/photo-1512221472476-7e439affc029?ixlib=rb-1.2.1&auto=format&fit=crop&w=1050&q=80",
    "price": {
        "cents": 0,
        "dollars": 85,
        "currency": "USD"
    },
    "name": "Airpods",
    "numPaid": 0,
    "numCompleted": 0,
    "description": "Cheap airpods suspendisse sed nisi lacus sed viverra tellus in hac habitasse platea dictumst vestibulum rhoncus est pellentesque elit ullamcorper dignissim cras tincidunt lobortis feugiat vivamus at augue eget arcu dictum.",
    "minCommits": 100,
    "deadline": "2020-07-08T15:59:00.000Z",
    "numCommits": 1,
    "oldPrice": {
        "currency": "USD",
        "cents": 0,
        "dollars": 100
    },
    "listingStatus": "ongoing",
    "id": 5081054809423872
}
```


server/src/api/listings.ts
```diff
@@ -37,4 +37,19 @@ listingRouter.get(
...
 );
 
+listingRouter.get(
+  '/:listingId',
+  async (req: Request, res: Response, next: NextFunction) => {
+    const {listingId: listingIdStr} = req.params;
+    const listingId = Number(listingIdStr);
+
+    try {
+      const listings = await listingService.getListing(listingId);
```
ifhuang, line 47: listing

server/src/api/listings.ts
```diff
@@ -37,4 +37,19 @@ listingRouter.get(
   }
 );
 
+listingRouter.get(
```
xlai20, line 40: Could we have a bit of documentation on the API functions? For example, let's say iIf I were a new user of this API, how shall I look up the right listingId in order to correctly call this API?

hzxcaryn:
@ifhuang Yea we should probably not send back numPaid and numCompleted if the one calling this API is not the merchant who posted the listing since it is not relevant to them. Will add an issue for this!

server/src/api/listings.ts
hzxcaryn: @xlai20 What do you mean by "look up the right listingId"? Also, where would you prefer this documentation to be? I was thinking maybe we can just edit our design doc section on API endpoints so we don't have to change it in many places.

xlai20: What defines as a listing id? Is it just a self-invented primary key?

The documentation should be in this file, or directly on top of the API method.

server/src/api/listings.ts
```diff
@@ -29,7 +28,22 @@ listingRouter.get(
   '/',
   async (req: Request, res: Response, next: NextFunction) => {
     try {
-      const listings = await listingService.getAllListings();
+      const listing = await listingService.getAllListings();
```
chshda, line 31: The original variable name (listings) sounds better.

hzxcaryn: @chshda Omg I am such a sotong, I edited the wrong variable 😭 🤣

server/src/api/listings.ts
hzxcaryn: @xlai20 Oh the listing id is the default primary key generated by datastore when we a new entity is created. Ok I understand what you want better now, will add it in!

Issue #58, opened 2020-06-22
[Onboarding] Customer can signin/signup


hzxcaryn:
Customers can sign in with their gpay accounts

PR #78, opened 2020-06-30
Add create customer endpoint

### LINKED ISSUE #71: Create add customer endpoint
Part of #58

hzxcaryn:
Part of #58 
Closes #71 

## Changes
- Added endpoint `POST /customers` to create a unique customer.
  * Accepts `gpayId`, `address` and `address` request body fields to create the customer. Only the `gpayId` field is mandatory.
  * Customers added must have unique `gpayId` field else the adding will fail.
  * A successful creation of the customer would return the details of the added customer in the response body (similar to a GET) with a 201 response code.
- Added a datastore wrapper for adding entities that ensures no duplicates on a unique property.
- Also realised that there was no need to dig so far in to extract the `id` of an inserted entity. We could just access `key.id` instead, which would be `undefined` if the transaction or insert operation failed, and the id of the inserted entity otherwise.
- Changed the `storage/datastore.ts` file to use inline exports instead because it was starting to look confusing which functions are exported members and which are not.

## TODO in future PRs (not in this PR)
- The same endpoint should try to "login" a customer. This means the endpoint would return customer information if the customer already exists with response code 200, before trying to add the customer (which is done in this PR)
- Validation of request body fields
- Proper sending of error codes
- [Authenticating of token](https://developers.google.com/identity/protocols/oauth2/openid-connect#validatinganidtoken)
- Tests

## How to test
Fetch branch
```
yarn
cd /server
yarn start:dev
```
Make a POST request to `http://localhost:5000/customers/` with at least a `gpayId` field.
Here's how a successful request look like:
image

If I try to make the same request again, I will get an error (wrong error code as of now):
image

And here is the added entity in the datastore:
image


server/src/storage/datastore.ts
```diff
@@ -64,35 +63,68 @@ const getAll = async (kind: string, filters?: Filter[]) => {
...
- * Unpacks ids of modified objects from CommitResponse object received from Datastore.
- */
-const getIdsFromCommitResponse = (
-  res: google.datastore.v1.ICommitResponse
-): (ResponseId | undefined)[] | undefined =>
-  res.mutationResults?.map(getIdFromMutationResult);
+  try {
+    await transaction.run();
+    const [queryRes] = await transaction.runQuery(query);
+    if (queryRes && queryRes.length > 0) {
```
ifhuang, line 86: queryRes?.

server/src/storage/datastore.ts
```diff
@@ -64,35 +63,68 @@ const getAll = async (kind: string, filters?: Filter[]) => {
...
+      // An entity with the unqiue property already exists
+      await transaction.rollback();
+      throw new Error(
```
ifhuang, line 89: this error should be caught outside, so you rollback twice?

server/src/storage/datastore.ts
```diff
@@ -64,35 +63,68 @@ const getAll = async (kind: string, filters?: Filter[]) => {
...
+ * @param uniqueProperty The property that should be unique for the specified kind
  */
-const getIdFromMutationResult = (
-  mutationResult: google.datastore.v1.IMutationResult
-): ResponseId | undefined => mutationResult?.key?.path?.[0]?.id;
+const insertUniqueEntity = async (
+  kind: string,
+  entity: Entity,
+  uniqueProperty: Filter
+): Promise<Entity | undefined> => {
```
ifhuang, line 77: it seems nothing is returned

server/src/storage/datastore.ts
```diff
@@ -64,35 +63,68 @@ const getAll = async (kind: string, filters?: Filter[]) => {
...
+    await transaction.rollback();
+    throw new Error(`Failed to add ${kind}. ${err}`);
+  }
+};
 
 /**
- * A Datastore wrapper that inserts a particular entity with the specified Kind.
+ * A Datastore wrapper that inserts a particular entity with the specified Kind and returns
+ * the id of the inserted entity.
+ * If uniqueProperty is specified, the entity would not be added if another entity with the same unique property value already exists.
```
ifhuang, line 106: line too long

server/src/services/customers.ts
```diff
@@ -14,10 +14,26 @@
...
+import {CustomerResponse, CustomerPayload} from '../interfaces';
 import {customerStorage} from '../storage';
 
 const getCustomer = async (customerId: number): Promise<CustomerResponse> =>
   await customerStorage.getCustomer(customerId);
 
-export default {getCustomer};
+/**
+ * Creates a customer if the customer is not already registered, and returns customer information.
+ * Returns existing customer information otherwise.
```
ifhuang, line 26: customerStorage.addCustomer should throw error if customer already exists right

server/src/api/customers.ts
```diff
@@ -39,4 +40,22 @@ customerRouter.get(
   }
 );
 
+/**
+ * Create a customer if they do not already exist.
+ */
+customerRouter.post(
```
xlai20, line 46: May I know why you name this folder as "API"? In my opinion, API sounds like a collection of function definitions waiting to be called by others.
But in this file, customerRouter.get and customerRouter.post are function called that get evaluated once when the file is imported.  They don't look like function definitions; correct me if I get it wrong.

server/src/api/customers.ts
```diff
@@ -39,4 +40,22 @@ customerRouter.get(
...
+  '/',
+  async (req: Request, res: Response, next: NextFunction) => {
```
xlai20, line 48: Another additional thing is that if you are already getting customer in the previous execution (customerRouter.get...), are you able to make judgement on whether the customer exists or not based on previous call? If so, you may avoid the addMerchant call if the customer is already known to exist.

server/src/storage/datastore.ts
```diff
@@ -64,35 +63,68 @@ const getAll = async (kind: string, filters?: Filter[]) => {
...
-): ResponseId | undefined => mutationResult?.key?.path?.[0]?.id;
+const insertUniqueEntity = async (
+  kind: string,
+  entity: Entity,
+  uniqueProperty: Filter
+): Promise<Entity | undefined> => {
+  const transaction = datastore.transaction();
+  const query = transaction
+    .createQuery(kind)
+    .filter(uniqueProperty.property, uniqueProperty.value);
```
xlai20, line 81: Could you explain why this implementation is doing insertion and throwing error if item exists, as the function doc specifies? Looks like it's just a query definition to me.

hzxcaryn: @xlai20 Yea this line is just creating the query. The insertion part is implemented in the try catch block from line 83 onwards

server/src/storage/datastore.ts
hzxcaryn: @ifhuang Oh okay I will remove this rollback and just leave the throwing of error here! Previously I wanted this to return something instead of throwing an error so I added the rollback call. But now that I have decided to throw an error instead, it is redundant as you have pointed out!

server/src/services/customers.ts
hzxcaryn: @ifhuang Oh yea! This was the previous behaviour and I forgot to change the docstring. Will change it!

server/src/storage/datastore.ts
hzxcaryn: You are right!

server/src/storage/datastore.ts
hzxcaryn: A second look made me realise that queryRes is of type `any[]`, so I can just do `queryRes.length > 0` check immediately!

server/src/api/customers.ts
hzxcaryn: @xlai20
> In my opinion, API sounds like a collection of function definitions waiting to be called by others.

This file is exactly what you think API is! :)

> But in this file, customerRouter.get and customerRouter.post are function called that get evaluated once when the file is imported. They don't look like function definitions; correct me if I get it wrong.

^ This is not right. These functions are not being called but these are how the router will handle `post` and `get` requests when it matches the route, which in this case is `/`(as stated in line 47), relative to `/customers/` (because this is a customerRouter, refer [here](https://github.com/googleinterns/gpay-group-buy/blob/master/server/src/index.ts#L36))
Syntax wise, this is just express syntax for routing, you can look at more examples from Express documentation [here](https://expressjs.com/en/4x/api.html#req.baseUrl).

server/src/api/customers.ts
hzxcaryn: @xlai20 As mentioned in the comment above, these are just specifying how the router will respond to these requests, and not executed. So this code block in here will execute independently of the get request.

server/src/api/customers.ts
```diff
@@ -39,4 +40,22 @@ customerRouter.get(
...
+    const customerData: CustomerPayload = req.body;
+
+    try {
+      const customer = await customerService.addCustomer(customerData);
+      res.status(201).send(customer);
```
karenfrilya97, line 53: You should also include 'Location' header in POST response.

> If a resource has been created on the origin server, the response SHOULD be 201 (Created) and contain an entity which describes the status of the request and refers to the new resource, and a Location header (see section 14.30).

from [here](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html).

Also, since you are returning the inserted object in the response body, should also include 'Content-Location' header, as mentioned in [this comment](https://stackoverflow.com/a/1832046) and the [definition of the header](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.14).

server/src/api/customers.ts
xlai20: I see. Then that's good then.

server/src/storage/datastore.ts
xlai20: Looks good then. The big block doc below gave me a wrong impression that that's another function definition.

server/src/api/customers.ts
hzxcaryn: @karenfrilya97 ok :D

server/src/storage/datastore.ts
```diff
@@ -64,35 +63,68 @@ const getAll = async (kind: string, filters?: Filter[]) => {
...
+    if (queryRes.length > 0) {
+      // An entity with the unqiue property already exists
+      throw new Error(
+        `Another entity with the same ${uniqueProperty.property} already exists.`
+      );
+    }
+    // Create the entity
+    transaction.insert(entity);
+    await transaction.commit();
+    return;
```
chshda, line 95: This is not needed?

server/src/services/customers.ts
```diff
@@ -14,10 +14,25 @@
...
+import {DEFAULT_CUSTOMER_PAYLOAD} from '../constants/default-payload';
+import {CustomerResponse, CustomerPayload} from '../interfaces';
 import {customerStorage} from '../storage';
 
 const getCustomer = async (customerId: number): Promise<CustomerResponse> =>
   await customerStorage.getCustomer(customerId);
 
-export default {getCustomer};
+/**
+ * Creates a customer if the customer is not already registered, and returns customer information.
```
chshda, line 25: Can we add more doc about what happens if the customer is already registered? Like does it throw exception?

server/src/storage/datastore.ts
hzxcaryn: That's true! 🤔

PR #82, opened 2020-06-30
Add merchant to database after signing up

### LINKED ISSUE #72: Send Merchant Sign Up form data to Add Merchant API endpoint

karenfrilya97:
Closes #72

Merge after #79 and #81


server/.env
```diff
@@ -0,0 +1,6 @@
+# These are environment variables for development in local machine.
```
hzxcaryn, line 1: If that's the case should it be `.env.development` or `.env.development.local` instead?

web/src/interfaces.ts
```diff
@@ -64,3 +64,22 @@ export interface Customer {
...
+ * would be sent to the server to add the merchant to the database.
+ */
+export interface MerchantPayload {
+  name: string;
+  email: string;
+  vpa: string;
+}
+
+/**
+ * MerchantResponse Interface that contains the fields of the Response that
```
xlai20, line 79: Doc change: ...Interface that contains the fields of the response that client side would receive after merchant signup.

web/src/interfaces.ts
```diff
@@ -64,3 +64,22 @@ export interface Customer {
...
+ * client side would receive.
+ */
+export interface MerchantResponse extends MerchantPayload {
```
xlai20, line 82: Probably the name "addMerchantResponse" is more appropriate to how you attempt to use it?

server/.env
karenfrilya97: Ohh cos dotenv cannot read `.env.development` :/

karenfrilya97: Okay actually dotenv-flow can read `.env.development`.

web/src/api.ts
```diff
@@ -33,3 +35,34 @@ export const getAllListings = async (): Promise<Listing[]> => {
...
+    `${process.env.REACT_APP_SERVER_URL}/merchants`,
+    requestOptions
+  );
+
+  if (res.status !== 201) {
+    throw new Error(Errors.SERVER_ERROR);
+  }
+
+  const location = res.headers.get('Location');
+  const merchantId = location?.substring(location?.lastIndexOf('/') + 1);
```
ifhuang, line 64: should we add an extra field to res to return merchant id?
extracting from location is not very intuitive

karenfrilya97: Yep! Actually that sounds good. I will make another PR for that?

web/src/interfaces.ts
karenfrilya97: It's not the response that client side would receive after merchant signup for now. But I changed it accordingly to make it clearer.

web/src/interfaces.ts
karenfrilya97: This response is supposed to be a representation for a Merchant object that is used in the HTTP response. So I think `MerchantResponse` would be more accurate.

PR #83, opened 2020-07-01
Add Merchant Sign In page

karenfrilya97:
Part of #2 

Changes made:
- Refactored out components of Sign Up that will be reused in Sign In.
- Added Sign In page using those components.

Differences from Merchant Sign Up page:
- No specific error message shown below each field, since there are only 2 fields.
- No checks for email format and password length, only if any field is empty.

image
The image above shows a screenshot of the Sign In page when a user has just filled in the email. The 'SIGN IN' button is disabled because there is an empty field.


web/src/components/merchant/sign-in/SignInCard.tsx
```diff
@@ -0,0 +1,60 @@
...
+ * This is the card containing the sign in form for the Merchant app.
+ * It is displayed in the Sign In page.
+ */
+const SignInCard: React.FC = () => {
+  return (
+    <CardContainer>
+      <GroupBuyMerchantHeader />
+      <SignInForm />
+      <StyledRow>
+        Already have an account? <Link to="/merchant/sign-up">Sign up</Link>{' '}
```
hzxcaryn, line 53: Do you think if we should use `&nbsp;`?

chshda, line 53: The string should be not having an account?

karenfrilya97: Oops, I forgot to change that! Thanks for pointing it out.

karenfrilya97:
> SignInCard, SignInForm have very similar styles as SignUpCard and SignUpForm. The PageContainer in `merchant/sign-in/index.tsx` as well.
> Maybe can extract them out as common components?

Yeah! Good idea. I should do that.

web/src/components/merchant/sign-in/SignInForm.tsx
```diff
@@ -0,0 +1,79 @@
...
+        label="Email"
+        inputType="email"
+        forwardedRef={validations.email}
+      />
+      <FormRow
+        label="Password"
+        inputType="password"
+        forwardedRef={validations.password}
+      />
+      <ErrorContainer></ErrorContainer>
```
ifhuang, line 71: no error msg?

web/src/components/merchant/sign-in/SignInCard.tsx
xlai20: +1. Can also add a reference back to sign in page from sign up page.

web/src/components/merchant/sign-in/SignInForm.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+import React from 'react';
+
+import FormRow from 'components/common/FormRow';
+import useSignInForm from 'components/merchant/sign-in/hooks/useSignInForm';
+import Button from 'muicss/lib/react/button';
+import Form from 'muicss/lib/react/form';
+import styled from 'styled-components';
+
+const StyledForm = styled(Form)`
```
xlai20, line 25: Can you re-use the styles between sign-in and sign-up forms? They look so similar; some styles look identical (e.g. StyledForm).

web/src/components/merchant/sign-in/SignInCard.tsx
karenfrilya97: There already is a reference from sign up page to sign in page. It's a similar link below the SIGN UP button.

web/src/components/merchant/sign-in/SignInForm.tsx
karenfrilya97: I haven't added that yet. From the current checks, the only errors are caused by empty email or password. But in that case, I will just disable the SIGN IN button. Will add error after I add the actual signing up process with firebase auth.

web/src/components/merchant/sign-in/SignInForm.tsx
karenfrilya97: Okay! Will refactor similar components out to reduce duplication.

web/src/components/merchant/sign-in/SignInCard.tsx
karenfrilya97: Is it for the `{' '}`? But `&nbsp;` means non-breaking space right, and I don't need it to be non-breaking.
Maybe I can just remove the `{' '}`?

web/src/components/merchant/sign-in/SignInForm.tsx
ifhuang: I'm ok with disable button for empty email or password.
But I thought this error msg is for server call (submit) failure?

karenfrilya97: Oh yepp, correct. Cos this PR hasn't done any server call yet.

PR #88, opened 2020-07-03
Make POST /customers endpoint return existing customer if customer with gpayId already exists

hzxcaryn:
Part of #58 

# Changes
- The `POST /customers` endpoint now returns the data of an existing customer if another customer with the same gpayId already exists. So now this is the updated behavior of `POST /customers`:
  * If a customer with the same gpayId exists, response body will include the details of the customer in the response body (similar to a get request), with status code 200
  * If the customer does not exists, this customer will be added to the datastore, and the reponse body will include the details of this newly added customer, with status code 201.

**Justification for using the same endpoint:**
We do not separate the endpoint because on the client side, we have no idea if a customer with the same gpayId already exists or not in our own database. Since we are using the Spot identity API to "login" users, we do not have separated "sign in" or "sign up" buttons, and hence we can just combine them into 1 endpoint to avoid extra round trip time to run both requests.
This behavior is actually similar to how Google OAuth "login" works, where a new account is created for new users, and if the user exists, we "login" instead.

# Testing

Here I will be running the `POST /customers` request twice with the same gpayId field. Note difference in the status code returned.

**Making the POST request for a new customer:**
Screenshot 2020-07-03 at 2 37 29 PM

**Making the POST request for an existing customer:**
Screenshot 2020-07-03 at 2 37 51 PM

### DIFF
--- a/server/src/api/customers.ts
+++ b/server/src/api/customers.ts
@@ -53,8 +53,10 @@ customerRouter.post(
     const customerData: CustomerPayload = req.body;
 
     try {
-      const existingCustomer = await customerService.getCustomerWithData(
-        customerData
+      // Customers are unique by their gpayId, so we will retrieve
+      // the customer using their gpay Id.
+      const existingCustomer = await customerService.getCustomerWithGpayId(
+        customerData.gpayId
       );
       if (existingCustomer !== null) {
         const resourceUrl = `${process.env.SERVER_URL}/customers/${existingCustomer.id}`;
--- a/server/src/api/customers.ts
+++ b/server/src/api/customers.ts
@@ -18,8 +18,6 @@
  * @fileoverview Handles routing of /customers endpoints.
  */
 
-import {readSync} from 'fs';
-
 import {Router, Request, Response, NextFunction} from 'express';
 import {CustomerPayload} from 'interfaces';
 
@@ -43,7 +41,11 @@ customerRouter.get(
 );
 
 /**
- * Create a customer if they do not already exist.
+ * Endpoint for the "logging in" of customers.
+ * If the customer exists, return the customer details in the
+ * response body with code 200.
+ * Else, add the customer and return the added customer details
+ * in the response body with code 201.
  */
 customerRouter.post(
   '/',
@@ -51,11 +53,21 @@ customerRouter.post(
     const customerData: CustomerPayload = req.body;
 
     try {
-      const customer = await customerService.addCustomer(customerData);
-      const resourceUrl = `${process.env.SERVER_URL}/customers/${customer.id}`;
+      const existingCustomer = await customerService.getCustomerWithData(
+        customerData
+      );
+      if (existingCustomer !== null) {
+        const resourceUrl = `${process.env.SERVER_URL}/customers/${existingCustomer.id}`;
+        res.setHeader('Content-Location', resourceUrl);
+        res.status(200).send(existingCustomer);
+        return;
+      }
+
+      const addedCustomer = await customerService.addCustomer(customerData);
+      const resourceUrl = `${process.env.SERVER_URL}/customers/${addedCustomer.id}`;
       res.setHeader('Content-Location', resourceUrl);
       res.location(resourceUrl);
-      res.status(201).send(customer);
+      res.status(201).send(addedCustomer);
       // TODO: Add error handling with the appropriate response codes.
     } catch (error) {
       return next(error);
--- a/server/src/interfaces.ts
+++ b/server/src/interfaces.ts
@@ -25,11 +25,18 @@ export interface CustomerPayload {
 }
 
 /**
+ * CustomerDatastoreReponse Interface that contains the fields of the Response that
+ * server side receives from the Datastore.
+ */
+export interface CustomerDatastoreReponse extends Required<CustomerPayload> {
+  id: number;
+}
+
+/**
  * CustomerResponse Interface that contains the fields of the Response that
  * client side would receive.
  */
-export interface CustomerResponse extends Required<CustomerPayload> {
-  id: number;
+export interface CustomerResponse extends CustomerDatastoreReponse {
   numOngoingCommits: number;
 }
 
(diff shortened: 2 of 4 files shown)


server/src/interfaces.ts
```diff
@@ -24,12 +24,19 @@ export interface CustomerPayload {
...
+  id: number;
+}
+
 /**
  * CustomerResponse Interface that contains the fields of the Response that
  * client side would receive.
  */
-export interface CustomerResponse extends Required<CustomerPayload> {
-  id: number;
+export interface CustomerResponse extends CustomerDatastoreReponse {
```
karenfrilya97, line 39: Why `extends CustomerDatastoreResponse` not `extends Required<...>`?

hzxcaryn: @karenfrilya97 Haha cos like all fields in CustomerDatastoreReponse is already required leh

karenfrilya97: but `id` is not required right?

server/src/services/customers.ts
```diff
@@ -21,6 +21,19 @@ import {customerStorage} from '../storage';
 const getCustomer = async (customerId: number): Promise<CustomerResponse> =>
   customerStorage.getCustomer(customerId);
 
+/**
+ * Gets a customer by identifying them with their data.
+ * Returns a CustomerResponse object when such a customer exits.
+ * Returns null otherwise.
+ * @param data Data of the customer that we are trying to get
+ */
+const getCustomerWithData = async (
```
karenfrilya97, line 30: Why not just have a `getCustomerWithGpayId`? I think it's weird for the caller of this method to pass in a whole `CustomerPayload` object when you just need one property. I'm guessing you don't want to expose the uniqueness of the gpayId to the api layer? But I think it's also weird because, what happens if the data doesn't include `gpayId`? Then, based on the name, you should find customer based on the remaining properties or something like that, which is not really done here.

server/src/interfaces.ts
hzxcaryn: What do you mean?  `CustomerDatastoreReponse` contains `id` as a non-optional field.
`Required<T>` is a TypeScript utility type that that makes all fields of type T required (aka non-optional). 
For example CustomerPayload includes optional fields
```
export interface CustomerPayload {
  gpayId: string;
  contactNumber?: string; // E164 format
  address?: string;
}
```
When I did 
```
export interface CustomerResponse extends Required<CustomerPayload> {
  id: number;
  numOngoingCommits: number;
}
```
It means `CustomerResponse` interface contains `contactNumber` and `address` as non-optional fields.

server/src/services/customers.ts
hzxcaryn: @karenfrilya97  Hmm yea I did think about that before. :( But I was thinking this is more of business logic that customers are unique by their gpayId instead of something that the api layer should be aware of :/ I'm not sure what's the best way around this either.

server/src/storage/customers.ts
```diff
@@ -30,10 +34,44 @@ const getCustomer = async (customerId: number): Promise<CustomerResponse> => {
...
+ * Returns a CustomerResponse object when such a customer exits.
+ * Returns null otherwise.
+ * @param gpayId The gpay id of the customer to retrieve
+ */
+const getCustomerWithGpayId = async (
+  gpayId: string
+): Promise<CustomerResponse | null> => {
+  const [customer]: CustomerDatastoreReponse[] = await getAll(CUSTOMER_KIND, [
+    {
+      property: 'gpayId',
```
ifhuang, line 63: Can datastore enforce uniqueness/index on gpayId?

server/src/services/customers.ts
ifhuang: +1 for getCustomerWithGpayId.
gpayId is required in CustomerPayload, so you'll always pass in gpayId. But I don't see future use cases by passing gpayId+other attribute (e.g. address) together

server/src/storage/customers.ts
hzxcaryn: @ifhuang Datastore is already indexed on gpayId! And uniqueness-wise datastore does not provide it so I'm the one implementing it in my `addCustomer` which uses a transaction to first ensure the same gpayId does not already exist.

ifhuang: SGTM

server/src/services/customers.ts
hzxcaryn: @ifhuang @karenfrilya97  Ok, changed

server/src/interfaces.ts
karenfrilya97: @hzxcaryn Okay never mind, I misunderstood the `Required<T>` utility type hahahahah :/ Thanks for explaining! Because Typescript doesn't enforce that the object contains the non-optional fields during runtime then I thought `Required<T>` would do that. But it doesn't. So you are right. Hahahah sorry for making you explain all this! But thanks for explaining anyway!! :D

server/src/api/customers.ts
```diff
@@ -43,19 +41,35 @@ customerRouter.get(
...
-      const customer = await customerService.addCustomer(customerData);
-      const resourceUrl = `${process.env.SERVER_URL}/customers/${customer.id}`;
+      // Customers are unique by their gpayId, so we will retrieve
+      // the customer using their gpay Id.
+      const existingCustomer = await customerService.getCustomerWithGpayId(
+        customerData.gpayId
+      );
+      if (existingCustomer !== null) {
+        const resourceUrl = `${process.env.SERVER_URL}/customers/${existingCustomer.id}`;
+        res.setHeader('Content-Location', resourceUrl);
```
karenfrilya97, line 63: Nice! You checked that this case doesn't need the 'Location' header but still requires the 'Content-Location'.

server/src/services/customers.ts
```diff
@@ -21,6 +21,17 @@ import {customerStorage} from '../storage';
 const getCustomer = async (customerId: number): Promise<CustomerResponse> =>
   customerStorage.getCustomer(customerId);
 
+/**
+ * Gets a customer by identifying them with their gpayId.
+ * Returns a CustomerResponse object when such a customer exits.
+ * Returns null otherwise.
+ * @param data Data of the customer that we are trying to get
```
chshda, line 28: This is not data any more

hzxcaryn: @chshda Good catch! 👍

Issue #87, opened 2020-07-03
Restrict GET /merchants request


karenfrilya97:
Restrict request such that the merchants themselves can get all the fields but other merchants/customers can get only public fields like name

PR #89, opened 2020-07-03
Create Get All Merchants API

karenfrilya97:
Add get merchants API that allows filters using query parameters.

**To do in future PRs:**
- (#87) Restrict response so that only the merchants themselves can get all the fields but other merchants/customers can get only public fields like name.
  - Right now, everyone can get all merchant objects including all the fields (name, email, VPA) except for password.

### DIFF
--- a/server/src/storage/merchants.ts
+++ b/server/src/storage/merchants.ts
@@ -20,11 +20,14 @@
  */
 
 import {MERCHANT_KIND} from '../constants/kinds';
-import {MerchantPayload} from '../interfaces';
-import {add} from './datastore';
+import {Filter, MerchantPayload, MerchantResponse} from '../interfaces';
+import {add, getAll} from './datastore';
 
 const addMerchant = async (merchant: MerchantPayload): Promise<number> =>
   add(MERCHANT_KIND, merchant);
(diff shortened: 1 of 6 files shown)


server/src/storage/merchants.ts
```diff
@@ -20,11 +20,15 @@
...
+import {Filter, MerchantPayload, MerchantResponse} from '../interfaces';
+import {add, getAll} from './datastore';
 
 const addMerchant = async (merchant: MerchantPayload): Promise<number> =>
   add(MERCHANT_KIND, merchant);
 // TODO(#67): Add checks to prevent adding multiple merchants with the same Firebase UID.
 
-export default {addMerchant};
+const getAllMerchants = async (
+  filters?: Filter[]
```
ifhuang, line 31: ideally we should also check the filters passed in, e.g. they should have valid property

hzxcaryn: @ifhuang Yep I think we should do field validation in the `services/` layer!

karenfrilya97: Yep! I will add a todo for that but don't think I should add that now based on the current timeline.

PR #103, opened 2020-07-07
Edit Merchant Auth middleware to just check token validity

karenfrilya97:
Changes:
- merchant-auth middleware now only checks if Firebase UID from decoded token matches the UID provided by the client, and throws 401 error otherwise.
- Firebase UID is now provided by the client instead of by the merchant-auth middleware as decoded token.

Reason for changes:
- Last time, I thought that Firebase UID was a secret that shouldn't be sent over HTTP request, but actually it's not.
- I realised that if merchant-auth provides the Firebase UID for the add merchant function, merchant-auth and Add Merchant endpoint becomes coupled, which is not ideal. Also, this defeats the single-responsibility principle since merchant-auth middleware also provides the content of one of the fields of the merchant to be added to the database.


server/src/middleware/merchant-auth.ts
```diff
@@ -29,13 +36,16 @@ const merchantAuth = async (
...
+    const {firebaseUid} = req.body;
+    if (firebaseUid !== undefined && firebaseUid !== verifiedFirebaseUid)
+      throw new Error('Invalid bearer token.');
+
     next();
-  } catch (error) {
-    res.sendStatus(401);
+  } catch (err) {
+    err.status = 401;
+    next(err);
```
xlai20, line 48: I'm curious on what the next function does? handle error?

karenfrilya97: @xlai20 `next` is an express parameter. Basically for express there are a series of functions called middlewares, handling the `request` and `response` objects. Then, `next` just means like this function is done and now it's passing on the responsibility to the next function in line. Here, the next function after this is an error handler which then handles the error.

For better explanation you can read with example here: https://expressjs.com/en/guide/using-middleware.html

PR #105, opened 2020-07-07
Create Add Listing API

### LINKED ISSUE #91: Create Add Listing API

karenfrilya97:
Closes #91 

### Features added:
- Add listing
- Merchant authentication:
  - `merchantId` in listing body must belong to the same merchant as the `firebaseUid` encrypted in the bearer token.

### Testing Efforts:
- Tried to add test but it fails because Datastore Emulator is outdated and doesn't allow ancestor-less query in transactions. So I will just include screenshots of the Postman tests I did.

**With valid token and body:**
image
image

**Without any token:**
image

**With expired token:**
image

**With token belonging to a different merchant from the listing's `merchantId`:**
image


server/src/services/listings.ts
```diff
@@ -14,14 +14,20 @@
...
-import {ListingResponse} from 'interfaces';
-
+import {DEFAULT_LISTING_PAYLOAD} from '../constants/default-payload';
+import {ListingRequest, ListingResponse} from '../interfaces';
 import {listingStorage} from '../storage';
 
+const addListing = async (listing: ListingRequest): Promise<ListingResponse> =>
+  listingStorage.addListing({
+    ...listing,
+    ...DEFAULT_LISTING_PAYLOAD,
```
hzxcaryn, line 24: I think just to be safe, you should swap the order of these 2 so that listing fields don't get overwritten by the default fields. (even though for this case we know that the `listing` won't contain the fields from the `DEFAULT_LISTING_PAYLOAD`)
```suggestion
    ...DEFAULT_LISTING_PAYLOAD,
    ...listing,
```

karenfrilya97: @hzxcaryn Ohh, thanks, I didn't know that! But, in this case, I actually want the listing fields to be overwritten by default fields because the fields like `numCommits` must be 0 at the point of creation, and similarly with the other fields in `DEFAULT_LISTING_PAYLOAD`.

hzxcaryn: @karenfrilya97 Oh yea, that makes sense for your use case of the default values! haha

server/src/middleware/merchant-auth.ts
```diff
@@ -29,13 +38,23 @@ const merchantAuth = async (
...
+    let firebaseUid;
+    if (req.body.firebaseUid !== undefined) {
+      firebaseUid = req.body.firebaseUid;
+    } else if (req.body.merchantId !== undefined) {
+      const merchant = await merchantService.getMerchant(req.body.merchantId);
+      firebaseUid = merchant.firebaseUid;
+    }
+
+    if (firebaseUid !== undefined && firebaseUid !== verifiedFirebaseUid)
+      throw new Error('Invalid bearer token.');
```
hzxcaryn, line 52: Would prefer braces! See https://google.github.io/styleguide/jsguide.html#formatting-braces-all

```suggestion
    if (firebaseUid !== undefined && firebaseUid !== verifiedFirebaseUid) {
      throw new Error('Invalid bearer token.');
    }
```

server/src/interfaces.ts
```diff
@@ -82,10 +82,21 @@ export type ListingStatus =
...
- * would be sent to create a Listing Entity.
+ * ListingComputedProperties Interface that contains the fields of a Listing that
+ * are not provided by the client, but computed by the server.
+ */
+export interface ListingComputedProperties {
+  numCommits?: number;
+  numPaid?: number;
+  numCompleted?: number;
+  listingStatus?: ListingStatus;
+}
```
hzxcaryn, line 93: Actually why did you make these fields optional?

karenfrilya97: @hzxcaryn more for future use case. LIke for example, if a listing passes its deadline, then the `listingStatus` would be changed but not the rest. Then can use this interface as a parameter for that function.

server/src/constants/default-payload.ts
```diff
@@ -22,3 +24,14 @@ export const DEFAULT_CUSTOMER_PAYLOAD = {
   contactNumber: '',
   address: '',
 };
+
+/**
+ * Contains the default values for the fields on a ListingPayload object which
+ * are automatically set by ListingService during listing creation.
+ */
+export const DEFAULT_LISTING_PAYLOAD: Required<ListingComputedProperties> = {
```
ifhuang, line 32: instead of creating this constant, can datastore support default value if not set?

server/src/middleware/merchant-auth.ts
karenfrilya97: @hzxcaryn Oh, okay! I was thinking of single-line arrow function where no braces is used. I thought the same style applies here. But I realised that case is different because without braces it allows implicit return which is different from with braces (i.e. not just style preference but has different semantics).

server/src/middleware/merchant-auth.ts
```diff
@@ -29,13 +38,23 @@ const merchantAuth = async (
...
+
+    let firebaseUid;
+    if (req.body.firebaseUid !== undefined) {
+      firebaseUid = req.body.firebaseUid;
+    } else if (req.body.merchantId !== undefined) {
+      const merchant = await merchantService.getMerchant(req.body.merchantId);
+      firebaseUid = merchant.firebaseUid;
+    }
+
+    if (firebaseUid !== undefined && firebaseUid !== verifiedFirebaseUid)
```
ifhuang: could firebaseUid be undefined?

server/src/interfaces.ts
hzxcaryn: @karenfrilya97 Ohh I see I see.

Do you think leaving it like this (aka `ListingComputedProperties` being fully optional, and using `Required<ListingComputedProperties>` when you need all of the fields to be non-optional) is better or changing it to something like

```
export interface ListingComputedProperties {
  numCommits: number;
  numPaid: number;
  numCompleted: number;
  listingStatus: ListingStatus;
}

export type ListingComputedPropertiesPartial = Partial<ListingComputedProperties>;
```
and then using the `ListingComputedPropertiesPartial` when you want all the fields to be optional?
Personally I like this more because it feels like "these are all the properties, but I just wanna update some of them, hence I use the partial version", but it might just be my own preference and it's up to you to decide!

server/src/api/listings.ts
```diff
@@ -55,4 +56,26 @@ listingRouter.get(
...
+    listingData.deadline = new Date(listingData.deadline);
+
+    try {
+      const listing = await listingService.addListing(listingData);
+      const resourceUrl = `${process.env.SERVER_URL}/listings/${listing.id}`;
+      res.setHeader('Content-Location', resourceUrl);
+      res.location(resourceUrl);
+      res.status(201);
+      res.json(listing);
+      // TODO: Add error handling with the appropriate response codes.
```
xlai20, line 74: Create an open issue to this TODO.

server/src/middleware/merchant-auth.ts
```diff
@@ -17,25 +17,46 @@
...
-    req.decoded = decodedFirebaseIdToken;
+    const verifiedFirebaseUid = await getVerifiedUid(firebaseIdToken);
+
+    let firebaseUid;
+    if (req.body.firebaseUid !== undefined) {
+      firebaseUid = req.body.firebaseUid;
+    } else if (req.body.merchantId !== undefined) {
+      const merchant = await merchantService.getMerchant(req.body.merchantId);
+      firebaseUid = merchant.firebaseUid;
+    }
```
xlai20, line 50: Can also check if firebaseUid is not defined (i.e. falling out of this if-elseIf block)?

server/src/constants/default-payload.ts
karenfrilya97: So far I haven't found anything that suggests that it supports default value yet. But I'm not too sure about that.
But either way, in our case, I prefer setting the values ourselves. For example, fields like listingStatus must have value 'ongoing'. I don't want to let the client provide another value for this field at the point of creation.

server/src/middleware/merchant-auth.ts
karenfrilya97: @xlai20 Yep, the check is in the next if block.

server/src/middleware/merchant-auth.ts
karenfrilya97: @ifhuang Yep, that's if neither of the 2 if-elseif conditions above are satisfied. It's related to @xlai20's comment above.

server/src/middleware/merchant-auth.ts
xlai20: You are referring to Line 52? That is not checking if firebaseUid is not defined right?
I'm thinking of (firebaseUid === undefined).

server/src/api/listings.ts
karenfrilya97: @xlai20 @hzxcaryn created the issue #115.

server/src/middleware/merchant-auth.ts
karenfrilya97: @xlai20 Ohh, I see! Sorry I thought you meant something else.
My idea was that merchant-auth prevents the client from 'pretending' to be another merchant. But if the request body doesn't contain any reference to any merchant (hence `firebaseUid === undefined`), then they are not 'pretending' to be someone they are not. So, there is no error thrown here.
I was thinking that if the field (`firebaseUid` or `merchantId`) is needed for whatever operation, then the error would be thrown in the main function instead (i.e. service or storage layer).

xlai20: I got your idea that (firebaseUid==undefined) is not very likely to happen and will be caught by the caller on service/storage layer.
However, as a way of defensive programming, the function here alone should not completely trust the data coming from its caller and still have to do some error handling (or at least an assertion) of a case falling out of if-else-if block. Sometimes a software runs into bug in mysterious way and these small assertion/data-validation code helps a lot in debugging, especially when your code gets altered by the future you or someone else in a way that the current you cannot foresee.

server/src/interfaces.ts
karenfrilya97: @hzxcaryn Ohh, that makes sense. Will do that! Thanks for the suggestion

server/src/middleware/merchant-auth.ts
karenfrilya97: @xlai20 Based on our discussion in chat, the problem seems to be that the if-elseif block looks buggy. So I will refactor out this block of code to remove the use of if-elseif with no else.

xlai20: Nice, I'll approve first to not block your submission. Make sure to fix it based on our discussion.

server/src/api/listings.ts
```diff
@@ -55,4 +56,26 @@ listingRouter.get(
   }
 );
 
+listingRouter.post(
+  '/',
```
chshda, line 60: So this path is used for both getting and adding listings?

karenfrilya97: @chshda Yep but they have different HTTP request methods.

PR #106, opened 2020-07-08
Create get all commits endpoint

### LINKED ISSUE #109: Add endpoint to get all commits
Part of #14 

Endpoint to call to get all commits.
Ability to filter retrieved commits via query params.

hzxcaryn:
Closes #109 

# Changes
- Add endpoint to get all commits that accept query parameters "listingId" and "customerId"
- Added tests for commits getter

**Not in this PR:**
-  Handling of proper error codes, added as a TODO.

# Testing
Invalid query param:
Screenshot 2020-07-08 at 2 53 54 PM

Invalid query value:
Screenshot 2020-07-08 at 2 54 29 PM

Non-existent id provided:
image

Successful case
Screenshot 2020-07-08 at 2 54 11 PM

Chaining query params
image

### DIFF
--- a/server/src/api/commits.ts
+++ b/server/src/api/commits.ts
@@ -36,7 +36,9 @@ commitRouter.get(
         res.sendStatus(400);
         return;
       }
-
+      // TODO: Parse queryParams json to make it CommitPayload type in runtime, which
+      // throws the appropriate type errors.
+      // Right now services is handling the type casting and throwing of errors.
       const commits = await commitService.getAllCommits(queryParams);
       res.status(200).json(commits);
       // TODO: Add error handling with the appropriate response codes.
--- a/server/src/api/commits.ts
+++ b/server/src/api/commits.ts
@@ -24,6 +24,9 @@ import {commitService} from '../services';
 
 const commitRouter = express.Router();
 
+/**
+ * Handles the get request for retrieving all commits.
+ */
 commitRouter.get(
   '/',
   async (req: Request, res: Response, next: NextFunction) => {
--- a/server/src/api/commits.ts
+++ b/server/src/api/commits.ts
@@ -30,6 +30,7 @@ commitRouter.get(
     try {
       const commits = await commitService.getAllCommits(req.query);
       res.status(200).json(commits);
+      // TODO: Add error handling with the appropriate response codes.
     } catch (error) {
       return next(error);
     }
--- a/server/src/api/commits.ts
+++ b/server/src/api/commits.ts
@@ -16,10 +16,24 @@
 
 /**
  * @fileoverview Handles routing of /commits endpoints.
- * @author Karen Frilya Celine
  */
 
-import express from 'express';
-const router: express.Router = express.Router();
+import express, {Request, Response, NextFunction} from 'express';
 
-export const commitRouter: express.Router = router;
+import {commitService} from '../services';
+
+const commitRouter = express.Router();
+
+commitRouter.get(
+  '/',
+  async (req: Request, res: Response, next: NextFunction) => {
+    try {
+      const commits = await commitService.getAllCommits(req.query);
+      res.status(200).json(commits);
+    } catch (error) {
+      return next(error);
+    }
+  }
+);
+
+export default commitRouter;
--- a/server/src/api/commits.ts
+++ b/server/src/api/commits.ts
@@ -31,7 +31,13 @@ commitRouter.get(
   '/',
   async (req: Request, res: Response, next: NextFunction) => {
     try {
-      const commits = await commitService.getAllCommits(req.query);
+      const queryParams = req.query;
+      if (Object.keys(queryParams).length === 0) {
+        res.sendStatus(400);
+        return;
+      }
+
+      const commits = await commitService.getAllCommits(queryParams);
       res.status(200).json(commits);
       // TODO: Add error handling with the appropriate response codes.
     } catch (error) {
--- a/server/src/index.ts
+++ b/server/src/index.ts
@@ -80,9 +80,13 @@ app.use(
 
 const localPort = process.env.NODE_ENV === 'test' ? 5001 : 5000;
 const port = process.env.PORT || localPort;
+
+// module.parent check to prevents the EADDRINUSE error for tests, following:
+// http://www.marcusoft.net/2015/10/eaddrinuse-when-watching-tests-with-mocha-and-supertest.html
 if (!module.parent) {
   app.listen(port, () => console.log(`Listening on port ${port}`));
 }
+
 const onStopSignal = () => {
   console.log('Termination signal received, exiting process..');
   process.exit(1); // eslint-disable-line no-process-exit
--- a/server/src/index.ts
+++ b/server/src/index.ts
@@ -80,8 +80,9 @@ app.use(
 
 const localPort = process.env.NODE_ENV === 'test' ? 5001 : 5000;
 const port = process.env.PORT || localPort;
-const server = app.listen(port, () => console.log(`Listening on port ${port}`));
-
+if (!module.parent) {
+  app.listen(port, () => console.log(`Listening on port ${port}`));
+}
 const onStopSignal = () => {
   console.log('Termination signal received, exiting process..');
   process.exit(1); // eslint-disable-line no-process-exit
@@ -93,4 +94,3 @@ process.on('SIGHUP', onStopSignal);
 process.on('SIGUSR2', onStopSignal); // Nodemon uses this
 
 export default app;
-export {server};
--- a/server/src/interfaces.ts
+++ b/server/src/interfaces.ts
@@ -62,6 +62,11 @@ export interface CommitPayload {
 }
 
 /**
+ * Union type of the keys of CommitPayload.
+ */
+export type CommitPayloadKey = keyof CommitPayload;
+
+/**
  * CommitResponse Interface that contains the fields of the Response that
  * client side would receive.
  */
@@ -140,4 +145,11 @@ export interface MerchantResponse extends MerchantPayload {
 export interface Filter {
   property: string;
   value: any;
+}
+
+/**
+ * QueryParams Interface that contains the fields of parameters in a query.
+ */
+export interface QueryParams {
+  [x: string]: any;
 }
--- a/server/src/interfaces.ts
+++ b/server/src/interfaces.ts
@@ -148,8 +148,8 @@ export interface Filter {
 }
 
 /**
- * QueryParams Interface that contains the fields of parameters in a query.
+ * A generic string key object.
  */
-export interface QueryParams {
-  [x: string]: any;
+export interface StringKeyObject {
+  [key: string]: any;
 }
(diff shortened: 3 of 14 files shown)


server/tests/integration/customers.test.ts
```diff
@@ -20,15 +20,14 @@ import app from '../../src';
 import customerFixtures from '../fixtures/customers';
 
 describe('Customers endpoints', () => {
-  test('it should fetch a single customer', async () => {
-    const expectedCustomerData = customerFixtures.data?.[0];
-    const customerId = customerFixtures.ids?.[0];
+  describe('GET /customers', () => {
+    test('it should fetch a single customer', async () => {
+      const expectedCustomerData = customerFixtures.responseData?.[0];
+      const customerId = customerFixtures.ids?.[0];
```
karenfrilya97, line 26: Why not this instead?
```suggestion
      const customerId = expectedCustomerData.id;
```
I think this can better ensure that the `id` is the correct id of the `expectedCustomerData`, and it is also immediately clear to the reader that the id belongs to the correct data.

hzxcaryn: Oh yea good catch! I was changing the customer one to match and didn't realise I should change that line too 😆

server/src/services/commits.ts
```diff
@@ -14,6 +14,51 @@
...
+ * Retrieves all commits.
+ * If query paramaters are provided, will retrieve all commits satisfying
+ * the query parameters.
+ * @param queryParams The query parameters
+ */
+const getAllCommits = async (
+  queryParams?: QueryParams
+): Promise<CommitResponse[]> => {
+  if (queryParams === undefined) {
+    return commitStorage.getAllCommits();
```
ifhuang, line 35: is there a valid use case for this?

server/src/services/commits.ts
```diff
@@ -14,6 +14,51 @@
...
+  }
+
+  const allowedKeys: CommitPayloadKey[] = ['customerId', 'listingId'];
```
ifhuang, line 38: use set instead of array

server/src/interfaces.ts
```diff
@@ -141,3 +146,10 @@ export interface Filter {
   property: string;
   value: any;
 }
+
+/**
+ * QueryParams Interface that contains the fields of parameters in a query.
+ */
+export interface QueryParams {
+  [x: string]: any;
```
ifhuang, line 154: call it 'key' instead of 'x'

server/src/index.ts
```diff
@@ -80,8 +80,9 @@ app.use(
 
 const localPort = process.env.NODE_ENV === 'test' ? 5001 : 5000;
 const port = process.env.PORT || localPort;
-const server = app.listen(port, () => console.log(`Listening on port ${port}`));
-
+if (!module.parent) {
```
ifhuang, line 83: what does this check do?

server/tests/integration/commits.test.ts
```diff
@@ -0,0 +1,98 @@
...
+ */
+
+import request from 'supertest';
+
+import app from '../../src';
+import commitFixtures from '../fixtures/commits';
+import customerFixtures from '../fixtures/customers';
+import listingFixtures from '../fixtures/listings';
+
+describe('Customers endpoints', () => {
```
ifhuang, line 24: Commits endpoints

server/src/index.ts
hzxcaryn: @ifhuang Haha I was trying to prevent the EADDRINUSE error for tests. http://www.marcusoft.net/2015/10/eaddrinuse-when-watching-tests-with-mocha-and-supertest.html

server/src/services/commits.ts
hzxcaryn: @ifhuang  yea I actually don't think there is a need to get all commits anymore HAHA shall return 404 instead if no query params are fed in

server/tests/integration/commits.test.ts
```diff
@@ -0,0 +1,95 @@
...
+      const targetCustomerId = customerFixtures.ids?.[0];
+      const targetListingId = listingFixtures.ids?.[0];
+      const expectedCommitData = commitFixtures.responseData.filter(
+        data =>
+          data.customerId === targetCustomerId &&
+          data.listingId === targetListingId
+      );
+
+      const res = await request(app).get('/commits').query({
+        customerId: targetCustomerId,
```
ifhuang, line 70: should also have listingId here

server/src/api/commits.ts
```diff
@@ -16,10 +16,34 @@
...
+    try {
+      const queryParams = req.query;
+      if (Object.keys(queryParams).length === 0) {
+        res.sendStatus(400);
+        return;
+      }
+
+      const commits = await commitService.getAllCommits(queryParams);
+      res.status(200).json(commits);
+      // TODO: Add error handling with the appropriate response codes.
```
xlai20, line 42: Do you have an open issue for this TODO? Is it dealing with the various response codes like "400""200" in this case?

server/src/api/commits.ts
```diff
@@ -16,10 +16,34 @@
 
 /**
  * @fileoverview Handles routing of /commits endpoints.
- * @author Karen Frilya Celine
```
xlai20, line 19: You probably don't want to delete the original author. Instead, add your own name into the author list.

server/src/index.ts
xlai20: Add your intention of this check as a comment next to it.

server/src/services/commits.ts
```diff
@@ -14,6 +14,50 @@
...
+const getAllCommits = async (
+  queryParams: StringKeyObject
+): Promise<CommitResponse[]> => {
+  const allowedKeys: Set<CommitPayloadKey> = new Set([
+    'customerId',
+    'listingId',
+  ]);
+  const filters: Filter[] = [];
+  Object.keys(queryParams).forEach(key => {
+    if (!(allowedKeys as Set<string>).has(key)) {
```
xlai20, line 40: Instead of having a separate if statement here and checking if key is in allowedKeys set, move everything into the switch statement below and let it become the default case.
```
switch (key) {
  case 'customerId':
  case 'listingId':
       value ...
   default:
       throw new Error(key is not valid query parameter)
 }
```
Then you don't need to create the const allowedKeys set at all.

server/tests/integration/commits.test.ts
```diff
@@ -0,0 +1,95 @@
...
+      expect(res.body).toEqual(expectedCommitData);
+    });
+
+    test('it should not filter by disallowed query params', async () => {
+      const res = await request(app).get('/commits').query({
+        createdAt: '2020-06-29T03:34:00.000Z',
+      });
+
+      expect(res.status).not.toBe(200);
+      expect(res.body).toHaveProperty('error');
```
xlai20, line 83: Can also test the error message details?

server/tests/integration/commits.test.ts
```diff
@@ -0,0 +1,95 @@
...
+    });
+
+    test('it should not filter by invalid query values', async () => {
+      const res = await request(app).get('/commits').query({
+        customerId: 'invalid-customer-id-type',
+      });
+
+      expect(res.status).not.toBe(200);
+      expect(res.body).toHaveProperty('error');
```
xlai20, line 92: Ditto. Error message details.

server/tests/integration/commits.test.ts
```diff
@@ -0,0 +1,95 @@
...
+
+describe('Commits endpoints', () => {
+  describe('GET /commits', () => {
+    test('it should not be able to fetch all commits', async () => {
+      const res = await request(app).get('/commits');
+
+      expect(res.status).toBe(400);
+    });
+
+    test('it should filter by listingId query param', async () => {
```
xlai20, line 32: Can you change the test description to:
either a complete sentence such as "It should filter by listingId query param." with capital first letter and dot at the end.
Or a verbal phrase "Should filter by listingId query param".

Ditto for the rest of test descriptions.

server/src/api/commits.ts
hzxcaryn: @xlai20 Oh I'm just removing this to standardise the files because if you check all other files, you will notice that we don't have the author tag anywhere else. This was added when Karen was creating the empty files during the setup.

server/src/services/commits.ts
hzxcaryn: @xlai20 Ohh that's a nice idea and would really simplify the existing code that I have here.

However, I was thinking of creating `allowedKeys` set so that we can be confident that the keys we allow would not contain keys that are not keys of `CommitPayload`. For example, I want to avoid stale code like the following to be left unchecked:
```
switch (key) {
  case 'customerId':
  case 'listingId':
  case 'oldKeyThatIsNoLongerPartOfCommitPayload':
       value ...
   default:
       throw new Error(key is not valid query parameter)
 }
```
Now, having `oldKeyThatIsNoLongerPartOfCommitPayload` in the query param would actually pass all the TS typechecks and not throw the error. Whereas if we have
```
  const allowedKeys: Set<CommitPayloadKey> = new Set([
    'customerId',
    'listingId',
    'oldKeyThatIsNoLongerPartOfCommitPayload',
  ]);
```
Then TS will alert us that `oldKeyThatIsNoLongerPartOfCommitPayload` is not part of `CommitPayloadKey`.

Actually I'm planning of adding the casting of query params json into a `CommitPayload` type at the `/api` layer before passing it to the `/services` layer, and when I do that, there would be no need for me to do the type casting in lines lines 46 to 51, and the entire switch block can be removed. This change to the `/api` layer would be done in a later PR (because typescript is only type-safe in compile time, so it is not a trivial change to cast query params into a `CommitPayload` type at the `/api` layer in runtime). That being said, I think I will just manually cast each field and throw an error in the `/api` layer for this PR. This way, the `/services` layer will just be in charge of the business logic of what specific keys are allowed as a query param, and not the parsing of input query params.

tldr; I will be moving out the type casting code to the api layer and keeping the `allowedKeys` in the services layer. (will request your review when the changes are made)

server/tests/integration/commits.test.ts
hzxcaryn: @ifhuang Oh good catch, didn't realise that I left that out! 😓

server/src/api/commits.ts
hzxcaryn: @xlai20 Nope not those, but rather dealing with those that are not returning a specific error code as of now. Okay will add an issue!

server/tests/integration/commits.test.ts
hzxcaryn: @xlai20 Ah, was thinking of adding those when I deal with all the specific error codes & error messages. But I guess I can just do a check for the error message for now!

server/tests/integration/commits.test.ts
hzxcaryn: @xlai20 Okay will change it to the latter!

server/src/services/commits.ts
xlai20: Noted. Remember to clean the line 46~51 if the type checking in allowedKeys is sufficient then.

PR #113, opened 2020-07-09
Create add commit endpt

### LINKED ISSUE #110: Add endpoint to add commits
Part of #11 
Endpoint to call when customer commits to a listing.
Remember to:
- [ ] Validate if customer is allowed to commit to another listing (whether they have already hit MAX_NUM_COMMITS or not)
- [ ] Use customer auth and check whether the customer is requesting for themselves
- [ ] Update commit count in listing as well during datastore transaction

hzxcaryn:
Closes #110 

# Changes
- Added endpoint to add commits
- Endpoint contains customer auth, so need a valid auth token to make the request
- Added `insertAndUpdateRelatedEntity()` transaction in the datastore layer  

**Not in this PR (to be done in future PRs):**
- Ensure that the user is making the request for themselves (can do this by checking the the decoded customer `sub` against the gpayId of the customer in the request.
- Proper parsing of req.body json to be of CommitRequest type
- Returning of proper error codes

# Testing
**Missing auth header**
Screenshot 2020-07-09 at 6 55 10 PM

**Using invalid listing or customer id**
Screenshot 2020-07-09 at 7 20 39 PM

**Customer has hit max commits**
Screenshot 2020-07-09 at 7 17 12 PM

**Successful request**
Screenshot 2020-07-09 at 6 52 38 PM

**Screenshot of datastore to show consistency between commit data and listing numCommits**
Screenshot 2020-07-09 at 6 53 46 PM
Screenshot 2020-07-09 at 6 53 28 PM

### DIFF
--- a/server/src/api/commits.ts
+++ b/server/src/api/commits.ts
@@ -19,7 +19,9 @@
  */
 
 import express, {Request, Response, NextFunction} from 'express';
+import {CommitRequest} from 'interfaces';
 
+import customerAuth from '../middleware/customer-auth';
 import {commitService} from '../services';
 
 const commitRouter = express.Router();
@@ -41,6 +43,29 @@ commitRouter.get(
       // Right now services is handling the type casting and throwing of errors.
       const commits = await commitService.getAllCommits(queryParams);
       res.status(200).json(commits);
+      // TODO: Add error handling with the appropriate response codes.
+    } catch (error) {
+      return next(error);
+    }
+  }
+);
+
+commitRouter.post(
+  '/',
+  customerAuth,
+  async (req: Request, res: Response, next: NextFunction) => {
+    // TODO: Parse req.body json to make it CommitRequest type
+    const commitData: CommitRequest = {
+      customerId: Number(req.body.customerId),
+      listingId: Number(req.body.listingId),
+    };
+
+    try {
+      const addedCommit = await commitService.addCommit(commitData);
+      const resourceUrl = `${process.env.SERVER_URL}/commits/${addedCommit.id}`;
+      res.setHeader('Content-Location', resourceUrl);
+      res.location(resourceUrl);
+      res.status(201).send(addedCommit);
       // TODO: Add error handling with the appropriate response codes.
     } catch (error) {
       return next(error);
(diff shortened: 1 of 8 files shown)


server/src/api/commits.ts
```diff
@@ -46,4 +48,31 @@ commitRouter.get(
...
+      listingId: Number(req.body.listingId),
+    };
+
+    try {
+      const addedCommit = await commitService.addCommit(commitData);
+      const resourceUrl = `${process.env.SERVER_URL}/commits/${addedCommit.id}`;
+      res.setHeader('Content-Location', resourceUrl);
+      res.location(resourceUrl);
+      res.status(201).send(addedCommit);
+      // TODO: Add error handling with the appropriate response codes.
```
xlai20, line 71: Create an open issue for the TODO.

server/src/services/commits.ts
```diff
@@ -60,4 +63,39 @@ const getAllCommits = async (
...
+/**
+ * Creates a new commit for the customer and the listing.
+ * An error would be thrown if the customer has reached their max num of ongoing commits,
+ * or if commit data provided is not valid.
+ * @param commitData Data of the commit to be added
+ */
+const addCommit = async (
+  commitData: CommitRequest
+): Promise<CommitResponse> => {
+  // Make sure that listing and customer exists
```
xlai20, line 75: Can there be a possibility that listing or customer does not exist and yet commit exists?

Is there a verification somewhere that ensure data consistency between COMMIT, LISTING, CUSTOMER and so on?

hzxcaryn: @xlai20 
> Can there be a possibility that listing or customer does not exist and yet commit exists?

Line 76 and 77 are the checks done to prevent the commit from being created if customer or the listing does not exits. 
The datastore layer uses transaction so updating of commit count in listing and the adding of a new commit would be atomic.

xlai20: Got it. Does your Line 76 and 77 throw exception when there is no listing or customer of the requested ID? Can you point me to the location where it will throw an exception?

hzxcaryn: @xlai20 Yep it does! The `datastore` layer throws an error when getting of a specified id fails. Can refer to the changes here: https://github.com/googleinterns/gpay-group-buy/pull/113/files#diff-f2fe6ca6032177087853d0c3ce45eedfR45-R51

PR #114, opened 2020-07-09
Created Empty Listings page UI for new merchant

karenfrilya97:
Screenshot 2020-07-10 at 3 15 18 AM - Display 1

**Changes:**
- Created UI skeleton for listings page, and empty listings placeholder.

**To-dos:**
- (#112) Make the UI more responsive.
- Show signed in merchant data instead of dummy data.
- Add checks to to show merchant's listings if they exist and show this page otherwise.

Part of #4, #6


web/src/components/common/GreenButton.tsx
```diff
@@ -0,0 +1,28 @@
...
+import styled from 'styled-components';
+
+import Button from './Button';
+
+/**
+ * Button with green background, white uppercase fonts and rounder corners.
+ */
+const GreenButton = styled(Button)`
+  background: var(--green);
+`;
```
hzxcaryn, line 26: Hmmmm for components like buttons I prefer that it's not a separate component on its own. Instead, you should make `Button` a general component and allow it to take in optional props of `color` where users can specify the color they want themselves. 

So you can do something like
```
<Button color="green">...</Button>
```
or allow user to fully customise the color they want 
```
<Button color="var(--green)">...</Button>
```
or you can do it the bootstrap-y way of accepting `"primary" | "danger"  | ...`  as the props etc

karenfrilya97: @hzxcaryn Oh yep! Was planning to do that also hahah

karenfrilya97:
> Just leaving a review first after doing a quick scan of the code!
> 
> Also, you don't have to export the icon svgs and place them in `assets/` because they are feather icons and we already have the dep in our project! Here [an example](https://github.com/googleinterns/gpay-group-buy/blob/34e6e0cdd93a357f3c0f87138dad096959d1d6f0/web/src/components/common/BackButton.tsx#L21) on how you can use them.
> 
> Oh and did not investigate yet but these buttons look misaligned (2nd button looks slightly more towards the right side)
> image

Oh thanks for those comments! :) Fixed them accordingly

web/src/components/merchant/sign-in/hooks/useSignInForm.tsx
```diff
@@ -59,7 +59,7 @@ const useSignInForm = () => {
       const {email, password} = values;
       await firebaseAuth.signInWithEmailAndPassword(email, password);
       const {id} = await getMerchantWithEmail(email);
```
karenfrilya97, line 61:
```suggestion
      await getMerchantWithEmail(email);
```

web/src/components/common/MerchantProfile.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+const DetailsContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  align-items: flex-start;
+  justify-content: center;
+`;
+
+/**
+ * This is a side bar for merchant pages. It shows merchant's profile and buttons
```
xlai20, line 41: This is a side bar? But it is named as "merchant profile"?

web/src/components/common/MerchantDetail.tsx
```diff
@@ -0,0 +1,66 @@
...
+interface MerchantDetailProps {
+  icon: React.FC<Props<AnyStyledComponent>>;
+  text: string;
+}
+
+/**
+ * This is a row that shows a field of the merchant details. It shows an icon
+ * describing the field and the text showing the value of the field.
+ */
+const MerchantDetail: React.FC<MerchantDetailProps> = ({icon, text}) => {
```
xlai20, line 52: Name this as MerchantDetailRow or MerchantDetailField

web/src/components/common/Button.tsx
```diff
@@ -0,0 +1,59 @@
...
+  'dark-gray': 'white',
+  green: 'white',
+  'pale-green': 'var(--green)',
+  'pale-gray': 'var(--gray)',
+};
+
+/**
+ * Button with dark gray background, white uppercase fonts and rounder corners.
+ */
+const Button = styled(MuiButton)`
```
chshda, line 31: Do we only have one kind of button? If not maybe give it a more concrete name

web/src/components/common/MerchantProfile.tsx
karenfrilya97: @xlai20 Oh! Thanks for noticing. I think I forgot to change the description.

web/src/components/common/MerchantDetail.tsx
karenfrilya97: @xlai20 Okay!

web/src/components/common/Button.tsx
karenfrilya97: @chshda Yep! So far there is only one generic button. But I guess we can name it something like `RoundedButton` to make it more specific.

web/src/assets/blank-canvas.svg
```diff
@@ -0,0 +1,120 @@
+<svg width="415" height="346" viewBox="0 0 415 346" fill="none" xmlns="http://www.w3.org/2000/svg">
```
ifhuang, line 1: is this svg only for merchant? if yes can we create a sub folder 'merchant'

web/src/components/merchant/listings/EmptyListingsPlaceholder.tsx
```diff
@@ -0,0 +1,62 @@
...
+const TextContainer = styled.div`
+  margin: 2% 0;
+  font-size: 1.5em;
+`;
+
+const StyledRoundedButton = styled(RoundedButton)`
+  font-size: 1.1em;
+`;
+
+const EmptyListingsPlaceholder: React.FC = () => {
```
ifhuang, line 40: add a comment for this

web/src/assets/blank-canvas.svg
karenfrilya97: @ifhuang okay!

PR #116, opened 2020-07-13
Add Merchant Context to pass signed-in merchant data to components

karenfrilya97:
Merge after #134 

### Changes:
- Add MerchantContext to 'store' signed-in merchant data and pass it to other pages after redirection.
  e.g. We can now store merchant data after sign in/sign up and then after redirecting, we can call merchant data from Home page.
- Store merchant data in MerchantContext after sign-in/sign-up.
- Retrieve merchant data from MerchantContext and display it in MerchantProfile component.

### Demo:
image
The merchant data above is no longer dummy data but data of currently signed-in merchant.

Screenshot 2020-07-17 at 1 45 47 PM
The picture above shows the page while the merchant data is being fetched after refreshing.


web/src/Routes.tsx
```diff
@@ -17,31 +17,19 @@
...
 import {Switch, Route} from 'react-router-dom';
 
-// Customer Pages
-const CustomerExplorePage = lazy(() => import('components/customer/explore'));
-// Merchant Pages
-const MerchantLandingPage = lazy(() => import('components/merchant/landing'));
-const MerchantSignUpPage = lazy(() => import('components/merchant/sign-up'));
-const MerchantSignInPage = lazy(() => import('components/merchant/sign-in'));
-// Common Pages
-const ListingDetailsPage = lazy(() => import('components/listing-details'));
```
ifhuang, line 30: Should this be common route?

web/src/components/merchant/contexts/MerchantContext.tsx
```diff
@@ -0,0 +1,93 @@
...
+};
+
+/**
+ * MerchantContext provider with merchant getter and setter functions as its value.
+ */
+const MerchantProvider: React.FC = ({children}) => {
+  const [merchant, setMerchant] = useState<MerchantType>(undefined);
+  const history = useHistory();
+
+  // Returns merchant stored in the context state if not undefined, otherwise
```
ifhuang, line 59: "if not undefined", can we just say if defined?

web/src/components/merchant/contexts/MerchantContext.tsx
```diff
@@ -0,0 +1,93 @@
...
+  // fetch merchant from server and update context state.
+  const getMerchant = async () => {
+    let result = merchant;
```
ifhuang, line 62: why not get rid of 'result' var?

web/src/components/merchant/Routes.tsx
```diff
@@ -0,0 +1,43 @@
...
+const MerchantOngoingListingsPage = lazy(() =>
+  import('components/merchant/ongoing-listings')
+);
+
+const MerchantRoutes: React.FC = () => {
+  const {path} = useRouteMatch();
+  return (
+    <Switch>
+      <MerchantProvider>
+        <Route exact path={path} component={MerchantLandingPage} />
```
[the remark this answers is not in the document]
karenfrilya97: @ifhuang So far it's nothing yet. We were thinking this could be a landing page to attract merchants to sign up, something like this https://cloud.google.com/

web/src/components/merchant/contexts/MerchantContext.tsx
karenfrilya97: @ifhuang I thought it's clearer because undefined has a specific meaning in TypeScript but defined might be slightly more vague.

web/src/components/merchant/contexts/MerchantContext.tsx
karenfrilya97: @ifhuang Oh ya I didn't realise I could just make it a `const` and declare it later.

web/src/Routes.tsx
karenfrilya97: @ifhuang Hmm I don't think it makes much of a difference but since we are talking about the components here I think 'page' is more accurate.

web/src/api.ts
```diff
@@ -81,23 +84,44 @@ export const getListing = async (listingId: number): Promise<Listing> => {
 };
 
 /**
- * Retrieves merchant with the given email from the database.
+ * Retrieves all merchants from the database.
```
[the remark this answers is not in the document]
karenfrilya97: @chshda okay!

web/src/Routes.tsx
xlai20: I think Yifu was asking about the ListingDetailsPage: Is this a common route shared between merchant and customer? If so, it should be moved into the customer routes alone; should have a separate Common Routes somewhere.

web/src/components/merchant/Routes.tsx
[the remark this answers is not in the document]
xlai20: Try not to create things or features that you are not building (or planing to build in the next 3 days) yet. We may end up with a delivery at the end of internship with quite some empty promises/Todos.

web/src/components/merchant/contexts/MerchantContext.tsx
```diff
@@ -0,0 +1,92 @@
...
+ * MerchantContext provider with merchant getter and setter functions as its value.
+ */
+const MerchantProvider: React.FC = ({children}) => {
+  const [merchant, setMerchant] = useState<MerchantType>(undefined);
+  const history = useHistory();
+
+  // Returns merchant stored in the context state if not undefined, otherwise
+  // fetch merchant from server and update context state.
+  const getMerchant = async () => {
+    if (merchant === undefined) {
```
xlai20, line 62: Use early return. 
```
if (merchant !== undefined) return merchant;

try {
 ...
...
```

web/src/components/merchant/contexts/MerchantContext.tsx
```diff
@@ -0,0 +1,92 @@
...
+      try {
+        const firebaseUid = await getFirebaseUid(); // Throws 'User not signed in' error.
+        const result = await getMerchantWithFirebaseUid(firebaseUid);
+        setMerchant(result);
```
xlai20, line 66: Your method title is called "getMerchant".
But in the code, your method does not seem to always about getting a merchant? (for example here you are setting merchant but not returning anything). The IF and ELSE flow are doing not semantically similar things. 

Consider either
1) Change the method name to reflect a universal meaning for all cases inside your method
or
2) Split them into two different methods, "getMerchant" and "UpdateMerchant".

web/src/components/merchant/Routes.tsx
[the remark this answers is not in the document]
hzxcaryn: @xlai20 @karenfrilya97 I think it is not nice to leave `/` as a `404` page either though! If you guys don't like the idea of an empty `MerchantLandingPage` I think for now the easiest thing we can do is smt like:
```suggestion
        <Route exact path={path}>
          <Redirect to=`${path}sign-in` />
        </Route>
```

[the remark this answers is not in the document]
hzxcaryn: @xlai20 @karenfrilya97 I think it is not nice to leave `/merchant` as a `404` page either though! If you guys don't like the idea of an empty `MerchantLandingPage` I think for now the easiest thing we can do is smt like:
```suggestion
        <Route exact path={path}>
          <Redirect to=`${path}sign-in` />
        </Route>
```

web/src/components/merchant/contexts/MerchantContext.tsx
karenfrilya97: @xlai20 Actually, after we `setMerchant(result)` we will leave the `if` block and `return result`. That's why I didn't use early return.
So, the flow is something like, if merchant is not defined yet, then we fetch it first, updateMerchant and then return, otherwise, just return straight away. I think if this is not clear I can refactor the content of the if block into a separate function.

web/src/Routes.tsx
karenfrilya97: @ifhuang @xlai20 @hzxcaryn To summarise what we discussed during today's meeting on this issue, previously listing route is common for both customer and merchant. If a customer/merchant goes to this route, they would be redirected to the customer/merchant's listing page respectively. However, for simplicity, we have decided not to have this route. There are now 2 separate routes for customer (/listing/:id) and merchant (/merchant/listing/:id).

web/src/components/merchant/Routes.tsx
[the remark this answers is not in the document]
karenfrilya97: @xlai20 Noted. This page was added as a part of the FE structure at the start and I just left it there.
@hzxcaryn I like that idea! Will add that in.

web/src/components/merchant/contexts/MerchantContext.tsx
karenfrilya97: @xlai20 I reply in the next comment (https://github.com/googleinterns/gpay-group-buy/pull/116#discussion_r455590863) because I feel like it's related.

web/src/components/common/MerchantProfile.tsx
```diff
@@ -40,20 +42,27 @@ const DetailsContainer = styled.div`
...
+    };
+
+    populateMerchantData();
+  }, [getMerchant]);
 
   return (
     <CentralisedContainer>
       <Shop />
-      <Header>{dummyMerchant.name}</Header>
+      <Header>{merchant?.name}</Header>
```
ifhuang, line 62: this prints null if merchant not yet fetched?
how about we move merchant null check to higher level, and only render content after merchant is fetched?

web/src/components/merchant/contexts/MerchantContext.tsx
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * MerchantContext provider with merchant getter and setter functions as its value.
+ */
+const MerchantProvider: React.FC = ({children}) => {
+  const [merchant, setMerchant] = useState<MerchantType>(undefined);
+  const history = useHistory();
+
+  // A helper method that fetches merchant data from database and updates
+  // merchant object stored in this context.
+  const fetchAndUpdateMerchant = async () => {
```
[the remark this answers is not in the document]
karenfrilya97: @ifhuang okay!

web/src/components/common/MerchantProfile.tsx
karenfrilya97: @ifhuang okay!

PR #122, opened 2020-07-13
Add delete commit endpoint

### LINKED ISSUE #111: Add endpoint to delete commits
Part of #12 

When customer uncommit from a listing

hzxcaryn:
Closes #111 

# Changes
- Added endpoint to delete commits
- Endpoint contains customer auth, so need a valid auth token to make the request
- Added `deleteAndUpdateRelatedEntity()` transaction in the datastore layer

**Not in this PR (to be done in future PRs):**
- Ensure that the user is making the request for themselves
- Returning of proper error codes

# Testing
**Deleting a non-existent Id**
image

**Deleting an invalid id**
Screenshot 2020-07-14 at 11 41 19 AM

**Deleting a non-ongoing listing**
Screenshot 2020-07-14 at 12 03 24 PM
Screenshot 2020-07-14 at 12 03 11 PM

**Successful delete**
Screenshot 2020-07-13 at 8 56 37 PM

**Before deleting in Datastore**
Screenshot 2020-07-13 at 8 56 04 PM

**After deleting in Datastore**
Screenshot 2020-07-13 at 8 56 59 PM

**Affected listing now has numCommits of 1 instead of 2**
Screenshot 2020-07-13 at 8 57 26 PM

### DIFF
--- a/server/src/api/commits.ts
+++ b/server/src/api/commits.ts
@@ -82,7 +82,7 @@ commitRouter.post(
  */
 commitRouter.delete(
   '/:commitId',
-  // customerAuth,
+  customerAuth,
   async (req: Request, res: Response, next: NextFunction) => {
     const {commitId: commitIdStr} = req.params;
     const commitId = Number(commitIdStr);
--- a/server/src/api/commits.ts
+++ b/server/src/api/commits.ts
@@ -77,4 +77,24 @@ commitRouter.post(
   }
 );
 
+/**
+ * Handles the delete requests to delete a commit with the specified commitId.
+ */
+commitRouter.delete(
+  '/:commitId',
+  // customerAuth,
+  async (req: Request, res: Response, next: NextFunction) => {
+    const {commitId: commitIdStr} = req.params;
+    const commitId = Number(commitIdStr);
+
+    try {
+      await commitService.deleteCommit(commitId);
+      res.sendStatus(204);
+      // TODO: Add error handling with the appropriate response codes.
+    } catch (error) {
+      return next(error);
+    }
+  }
+);
+
 export default commitRouter;
--- a/server/src/api/commits.ts
+++ b/server/src/api/commits.ts
@@ -88,6 +88,10 @@ commitRouter.delete(
     const commitId = Number(commitIdStr);
 
     try {
+      if (Number.isNaN(commitId)) {
+        throw new Error('Invalid commitId params.');
+      }
+
       await commitService.deleteCommit(commitId);
       res.sendStatus(204);
       // TODO: Add error handling with the appropriate response codes.
--- a/server/src/services/commits.ts
+++ b/server/src/services/commits.ts
@@ -103,14 +103,19 @@ const addCommit = async (
 
 /**
  * Deletes the commit with the specified commitId.
- * An error would be thrown if commit does not already exist, and if deletion fails.
+ * An error would be thrown if commit does not already exist,
+ * or if the commit is not ONGOING,
+ * or if deletion fails.
  * @param commitId Id of the commit to be deleted
  */
 const deleteCommit = async (commitId: number) => {
   const commit = await commitStorage.getCommit(commitId);
-  const commitListing = await listingStorage.getListing(commit.listingId);
 
-  return commitStorage.deleteCommit(commitId, commitListing.id);
+  if (commit.commitStatus !== 'ongoing') {
+    throw new Error('Only ongoing commits are allowed to be deleted.');
+  }
+
+  return commitStorage.deleteCommit(commitId, commit.listingId);
 };
 
 export default {getAllCommits, addCommit, deleteCommit};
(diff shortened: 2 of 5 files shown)


server/src/storage/datastore.ts
```diff
@@ -135,6 +135,45 @@ export const insertAndUpdateRelatedEntity = async (
   return Number(keyToInsert.id);
 };
 
+/**
+ * Deletes an entity and update a related entity in the same transaction.
```
ifhuang, line 139: "updates"

server/src/services/commits.ts
```diff
@@ -94,11 +94,23 @@ const addCommit = async (
...
     ...DEFAULT_COMMIT_PAYLOAD,
     ...commitData,
     createdAt: new Date(),
   });
 };
 
-export default {getAllCommits, addCommit};
+/**
+ * Deletes the commit with the specified commitId.
+ * An error would be thrown if commit does not already exist, and if deletion fails.
```
ifhuang, line 106: "or if deletion fails"

server/src/services/commits.ts
```diff
@@ -94,11 +94,23 @@ const addCommit = async (
...
+ * @param commitId Id of the commit to be deleted
+ */
+const deleteCommit = async (commitId: number) => {
+  const commit = await commitStorage.getCommit(commitId);
```
ifhuang, line 110: You should have check on commit status right? e.g. can you delete a paid commit?

server/src/services/commits.ts
```diff
@@ -94,11 +94,23 @@ const addCommit = async (
...
+  const commitListing = await listingStorage.getListing(commit.listingId);
+
+  return commitStorage.deleteCommit(commitId, commitListing.id);
```
karenfrilya97, line 113: Why do you need to retrieve the listing object? You could have just done this?
```suggestion
  return commitStorage.deleteCommit(commitId, commit.listingId);
```

karenfrilya97, line 113: Why do you have to get the listing id in this layer? This feels more like a storage thing than a logic thing. It's not really a business logic that require you to retrieve the commit object and get the related listing id, but more related to the way these data are stored.

server/src/services/commits.ts
hzxcaryn: @ifhuang Yes I should check if it's ongoing! Oops was gonna do that but I got distracted by the datastore method and it slipped my mind, sorry! 😅

server/src/services/commits.ts
hzxcaryn: @karenfrilya97 Yep you are right! Haha I forgot that I was halfway through implementing the services layer (I was gonna do some validation against listing but now I realised I don't have to) when I got distracted by implementing the datastore layer and it slipped my mind to come back to it! 🙈 🙈 Sorry for the confusion!

server/src/api/commits.ts
```diff
@@ -77,4 +77,24 @@ commitRouter.post(
   }
 );
 
+/**
+ * Handles the delete requests to delete a commit with the specified commitId.
+ */
+commitRouter.delete(
+  '/:commitId',
+  // customerAuth,
```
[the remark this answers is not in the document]
hzxcaryn: @chshda Oops I forgot to uncomment 😭  Thanks for catching this!

server/src/api/commits.ts
```diff
@@ -77,4 +77,24 @@ commitRouter.post(
...
+  async (req: Request, res: Response, next: NextFunction) => {
+    const {commitId: commitIdStr} = req.params;
+    const commitId = Number(commitIdStr);
```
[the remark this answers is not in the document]
hzxcaryn: @chshda Oh actually was just thinking of dealing with the happy path similar to the customers getter [here](https://github.com/googleinterns/gpay-group-buy/blob/master/server/src/api/customers.ts#L32). But you are right, not dealing with it now/not including a TODO will risk this issue to be overlooked in the future. 
I will add the check for both commits and the customer getter to be consistent!

server/src/services/commits.ts
hzxcaryn: @karenfrilya97 Hmm I agree it felt abit odd passing both commitId and listingId in as arguments, but I also don't really want to fetch commit twice, both in `services` layer and in `storage` layer. (I need to get in `services` layer because I want to do biz logic checks on commitStatus) One way is to pass the `commit` object to the `storage` layer, but that's weird too because this `commit` object is a `CommitResponse` object

karenfrilya97: @hzxcaryn Ohh hahahah no wonder! That happens to me, too 😄

karenfrilya97: @hzxcaryn Ohh! I didn't know it's for `commitStatus` checking. In that case that's fine!
I personally think passing `commit` object looks nice (because it's something like 'please delete this commit'), but I also don't like the idea of passing more than what is necessary. I guess this is fine for now. Maybe if we can think of a better way to do it in the future we can change it.

PR #124, opened 2020-07-15
Add commit uncommit initial flow

### LINKED ISSUE #11: Customer can commit to a listing
**Mockup:** https://www.figma.com/file/x0PWXRWxwSopiixYv41mK7/GPay-GroupBuy?node-id=1%3A41681

### LINKED ISSUE #123: Customer can uncommit from a listing from listing details page
Part of #12 

https://www.figma.com/file/x0PWXRWxwSopiixYv41mK7/GPay-GroupBuy?node-id=1%3A42273

hzxcaryn:
Closes #11, Closes #123

# Changes
- Added button to commit/uncommit from a listing (depending on the commit status) in the listing-details page
- Added `ListingDetailsContext` with `useListingDetailsContext` hook to separate out the logic of commits and listing of the customer & the listing
  * Currently, I am using hardcoded dummy customerId and token, because I don't want to make the changeset of this PR too huge, will fix in next PR!
- Added helper functions in the `api` layer so that we don't have to repeat the same headers everytime (I felt it was getting repetitive haha)

**Not in this PR, TODO for future PRs**
- Remove temporary dummy customerId and token used in `useCommitStatus` after storing these details in a CustomerContext
- UI/UX dealing with the un-happy paths 

## Screenshots

**Transition from before clicking the commit button to after clicking the commit button**
Screenshot 2020-07-14 at 10 23 46 PM
Screenshot 2020-07-14 at 10 26 14 PM
Screenshot 2020-07-14 at 10 23 34 PM
Clicking the 'UNCOMMIT' button will just be these screenshots in reversal hahaha

# How to test

If you want to try it out for yourself, it might be a bit troublesome as you would need to generate a valid token from the microapp identity API. You can do that by connecting your device running the microapp and inspecting the network packet after clicking the "GET IDENTITY" button at the bottom of the Explore page for the JWT. Alternatively you can ask me for a valid token! 

After that, you can fetch my branch, replace `const token = 'replace-with-a-valid-token';` with the valid token and try it out locally!


web/src/components/customer/listing-details/ActionBar.tsx
```diff
@@ -0,0 +1,117 @@
...
+    (buttonState: ActionButtonState): JSX.Element => {
+      const onClickButton = async (callback: () => Promise<void>) => {
+        setButtonState('loading');
+        await callback();
+      };
+
+      switch (buttonState) {
+        case 'loading':
+          return <ActionButton disabled>Loading...</ActionButton>;
+        case undefined:
```
xlai20, line 79: Move this undefined case below all the named cases.

And actually, why do you have an "undefined" button state in the first place? Can you use "initial" as a more explicit name?

hzxcaryn: @xlai20 Oh because I was extending on the types of CommitStatus (for example, when there is no commit found for this customer + listing, then we get undefined) But yep I should set button state to `initial` if CommitStatus is undefined!

ifhuang:
Looking at your second screenshot: commit button is loading, but commit progress bar already increased 1.
Should we increase commit progress bar after the commit button loading state finished?

web/src/components/customer/listing-details/ActionBar.tsx
```diff
@@ -0,0 +1,114 @@
...
+          return (
+            <ActionButton
+              onClick={() => onClickButton(onUncommit)}
+              color="danger"
+            >
+              Uncommit listing
+            </ActionButton>
+          );
+        default:
+          // TODO: Deal with other cases, such as paid & successful cases
```
ifhuang, line 92: We'd better to have a state: cannot commit due to max commit num reached

hzxcaryn: @ifhuang Omg I can't believe I forgot about that! Thanks for pointing that out, I will add that case into the TODO comment!

hzxcaryn:
@ifhuang 
> Looking at your second screenshot: commit button is loading, but commit progress bar already increased 1.
Should we increase commit progress bar after the commit button loading state finished?

Oh actually it only updates after the `commitStatus` is changed, because it reacts on `commitStatus` change! 
I think when I was taking the screenshot, I was not fast enough and was unable to catch the "Loading..." status of the button when I click commit so I clicked it again (hence uncommitting) and managed to take the screenshot. HAHA

PR #128, opened 2020-07-15
Add mobile prompt UI component

hzxcaryn:
# Changes
- Added a mobile prompt reusable UI component 
  * Will be needed in the coming PRs for pages like https://www.figma.com/file/x0PWXRWxwSopiixYv41mK7/GPay-GroupBuy?node-id=28%3A851 
- Added a demo of how it can be used + how it looks like in `/design-samples`

## Screenshots
Screenshot 2020-07-15 at 8 06 02 PM
Screenshot 2020-07-15 at 8 06 09 PM

# How to test
Fetch into my branch,
```
cd web
yarn 
yarn start
```
and navigate to [localhost:3000/design-samples](http://localhost:3000/design-samples) to try it out!


web/src/components/common/MobilePrompt.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+  font-weight: bold;
+
+  && {
+    color: var(--green);
+  }
+`;
+
+/* Prevent body scroll code from https://css-tricks.com/prevent-page-scrolling-when-a-modal-is-open/ */
+const preventBodyScroll = () => {
```
chshda, line 97: nit: I think disableBodyScroll pairs better with enableBodyScroll below.

web/src/components/common/MobilePrompt.tsx
```diff
@@ -0,0 +1,162 @@
...
+}) => {
+  useEffect(() => {
+    if (isVisible) {
+      preventBodyScroll();
+    } else {
+      enableBodyScroll();
+    }
+  }, [isVisible]);
+
+  useEffect(() => () => enableBodyScroll(), []);
```
chshda, line 140: Do we need two "() =>"?

web/src/components/common/MobilePrompt.tsx
hzxcaryn: @chshda Oh nice suggestion!

web/src/components/common/MobilePrompt.tsx
hzxcaryn: @chshda Nice catch, we don't!

web/src/components/common/MobilePrompt.tsx
```diff
@@ -0,0 +1,162 @@
...
+/* End prevent body scroll code from https://css-tricks.com/prevent-page-scrolling-when-a-modal-is-open/ */
+
+interface ButtonDetails {
+  name: string;
+  onClick: () => void;
+}
+
+interface MobilePromptProps {
+  isVisible: boolean;
+  title?: string;
```
ifhuang, line 117: title could be optional?

web/src/components/common/MobilePrompt.tsx
```diff
@@ -0,0 +1,162 @@
...
+  header?: JSX.Element;
+  buttons?: ButtonDetails[];
```
ifhuang, line 119: dismiss button is not required, so user can dismiss by clicking the background?

hzxcaryn: @ifhuang Hmm nope they cannot! 
In https://www.figma.com/file/x0PWXRWxwSopiixYv41mK7/GPay-GroupBuy?node-id=107%3A727 we have a "Confirm" button instead of a Dismiss button, so I wanted to make the buttons customisable.
Would you prefer if I checked at at least 1 button is present in buttons?

web/src/components/common/MobilePrompt.tsx
hzxcaryn: @ifhuang I can make it required! I don't think we have a usecase for a title-less prompt actually

ifhuang: sg, let's make it required then

web/src/components/common/MobilePrompt.tsx
ifhuang: > Would you prefer if I checked at at least 1 button is present in buttons?

ok.

Can you also add comments about how can the prompt be dismissed to the MobilePrompt class

PR #131, opened 2020-07-16
Add check for listingStatus on add commit endpoint

hzxcaryn:
# Changes 
- Previously, I left out the check that listing must be ongoing before allowing commit to be added. This PR fixes this.
- Also realised that I can simplify the check for customer's numOngoingCommits since I have already fetched the customer!
# Testing efforts
**Check that we can't commit to non-ongoing listings**
Screenshot 2020-07-16 at 1 44 23 PM

**Check that checking for max num commits still works**
Screenshot 2020-07-16 at 1 45 05 PM


server/src/services/commits.ts
```diff
@@ -74,21 +74,18 @@ const addCommit = async (
   commitData: CommitRequest
 ): Promise<CommitResponse> => {
   // Make sure that listing and customer exists
```
karenfrilya97, line 76: Not very important but can I think it's better to change this comment to `// Check ...` to standardise with the other check comments.
Also, to standardise, should add full stop after the comment.

hzxcaryn: @karenfrilya97 Ok will change to "Check"! But for fullstop wise I think normally our `// ...` comments don't have (is more like a phrase) but `/** ... */` comments have

PR #134, opened 2020-07-16
Separate Customer and Merchant routes into nested routes

karenfrilya97:
Previously, all routes are in /web/src/Routes.tsx. There are 2 problems with this:
- The list of routes may become too long and hard to read.
- We cannot have separate Customer and Merchant contexts for Customer and Merchant routes respectively, since each `<Switch>` component can only contain 1 context provide as its child.

This PR separates these routes into nested Customer and Merchant routes.


web/src/Routes.tsx
```diff
@@ -17,33 +17,19 @@
...
-const MerchantListingsPage = lazy(() => import('components/merchant/listings'));
-// Common Pages
-const ListingDetailsPage = lazy(() => import('components/listing-details'));
 // Design samples
 const DesignSamplesPage = lazy(() => import('components/design-samples'));
 
 const Routes: React.FC = () => (
   <Suspense fallback={<Loading />}>
     <Switch>
-      <Route exact path="/merchant" component={MerchantLandingPage} />
```
ifhuang, line 38: I see you removed this landing page, is it expected?

karenfrilya97: @ifhuang based on this comment (https://github.com/googleinterns/gpay-group-buy/pull/116#discussion_r453729049), I decided to just remove it for now. I will add that in later in another PR according to Caryn's suggestion.

karenfrilya97:
> continuous-integration/travis-ci/pr seems to be failing. pls fix

It's fixed now! It's weird because the backend is failing but this PR doesn't make any changes to the backend :/

web/src/components/customer/Routes.tsx
```diff
@@ -0,0 +1,41 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React, {lazy} from 'react';
+
+import CustomerCommitCountProvider from 'components/customer/contexts/CommitCountContext';
+import {useRouteMatch, Switch, Route} from 'react-router-dom';
+
+// Customer Pages
```
hzxcaryn, line 22: I think we don't need this comments anymore? :p

PR #135, opened 2020-07-16
Add FormPropsContext to pass form props to nested children components

karenfrilya97:
### Changes: (Refactoring with no behaviour/UI change)
- Add FormPropsContext.
- Component which has a Form can pass the props to the context.
- Form components (which may be nested multiple levels deep) can get the props from the context.
- Intermediate components don't need to bother with passing on the props.


karenfrilya97:
> Have not looked thoroughly but just wanted to bring this point up!
> How is this different from the one that comes together with react-hook-form https://react-hook-form.com/api/#useFormContext?

Yep, I saw that, too. I was considering to use this as well. But some of the props to be passed on are custom data like fields and validations which are not from react-hook-form.

PR #137, opened 2020-07-17
Add hook to handle Add Listing form submission

karenfrilya97:
### Changes
- Add logic to handle Add Listing form submission which should do the following:
  - Format values received from Add Listing form into ListingPayload interface.
  - Send listing data to Add Listing endpoint in the server.
  - Redirect user to Ongoing Listings page on success.
- Add `parseMoney` function to parse the number input and currency string from Add Listing form.

### Testing
image
Added a button on Add Listing page, which when clicked, sends a POST request to /listings endpoint in the server, containing dummy listing data in the body.


web/src/api.ts
```diff
@@ -178,23 +198,36 @@ export const deleteCommit = async (commitId: number, idToken: string) => {
...
+export const getMerchantWithFirebaseUid = async (
+  firebaseUid: string
+): Promise<MerchantResponse> => {
+  const merchants = await getAllMerchants({
+    firebaseUid,
+  });
 
   if (merchants.length === 0) {
-    throw new Error(USER_NOT_FOUND);
+    throw new Error(NO_MERCHANT_WITH_FIREBASE_UID);
```
xlai20, line 230: Include the Firebase UID that is throwing error in this error message as well.

web/src/utils/money/index.ts
```diff
@@ -26,3 +26,14 @@ export const formatMoney = ({currency, dollars, cents}: Money): string => {
     amount
   );
 };
+
+/**
+ * Formats value number and currency string into Money type.
+ * @param number The value of the Money in number type
+ * @param currency The 3-letter currency code defined in ISO 4217
+ */
+export const parseMoney = (value: number, currency: string): Money => ({
```
chshda, line 35: Will be good to document the behaviour about the precision.

web/src/api.ts
karenfrilya97: @xlai20 okay!

web/src/utils/money/index.ts
karenfrilya97: @chshda okay!

PR #138, opened 2020-07-17
Add customer context

hzxcaryn:
To be merged after #130 

# Changes
- Previously, we have a `CommitCountContext` to provide commit count of a customer. Now, I have changed to use a `CustomerContext` to provide stateful customer data. Now all customer login logic is contained inside the `CustomerContext` as well.
- Removed hardcoded sample `customerId`s and `idToken`s. After this PR is in, commit uncommit flow (happy path) would work properly in the gpay app :D


web/src/components/customer/listing-details/contexts/ListingDetailsContext.tsx
```diff
@@ -75,29 +73,44 @@ const ListingDetailsProvider: React.FC<ListingDetailsProviderProps> = ({
...
         setCommitId(commit.id);
         setCommitStatus(commit.commitStatus);
       }
     };
     fetchCommit();
-  }, [listingId]);
+  }, [listingId, customer]);
 
   const onCommit = async () => {
+    if (customer === undefined || idToken === undefined) {
```
ifhuang, line 95: can we just check customer?
is there a case that customer defined but idToken undefined?

hzxcaryn: @ifhuang Logically yes can just check customer! But then we can't just do that because typescript will complain about `idToken` possibly being undefined hahaha

PR #141, opened 2020-07-20
Show merchant listings on Listings page

karenfrilya97:
Merge after #142.

### Changes
- [UI] Show merchant listings if merchant has any, and show empty listings placeholder otherwise.
- [Refactoring] Move ListingCollection component from components/customer/explore/ folder to components/common and reuse it for merchant listings page.
- [Debugging] Made some minor changes in MerchantContext to fix a bug.

### Test
**Listings page of merchant with listings**
image

**Listings page of merchant without listings**
image

### Todo
- Split listings into Ongoing and Past listings.

### Linked Issues
Part of #4, #6.


web/src/components/common/ListingCollection.tsx
```diff
@@ -44,44 +44,41 @@ const ListingItem: React.FC<ListingProps> = ({
...
-      }}
-    >
-      <ListingCard
-        listingName={name}
-        price={price}
-        oldPrice={oldPrice}
-        endDate={deadline}
-        imgUrl={imgUrl}
+}) => {
+  const isMobile = useMediaQuery({query: '(max-device-width: 768px)'});
```
ifhuang, line 48: can you move this to a common place so that it could be used by other places?
also this check looks not straightforward, is there other way like checking the user agent?

hzxcaryn, line 48: Would prefer values like `768px` to be in a new `constants/media-query.ts` file instead! It can probably be called `SM_MIN`? 

Here is the bootstrap grid system, and we can probably use their sizes as a guide for our constants file :D  
```
xs (for phones - screens <768px wide)
sm (for tablets - screens =>768px wide)
md (for small laptops - screens =>992px wide)
lg (for laptops and desktops - screens =>1200px wide)
```
These constants can then be used in `useMediaQuery`s and also in `@media` in our css styling.

karenfrilya97: @ifhuang Okay! I will put this in a context like @hzxcaryn suggested. 
Hmm I don't know of a more straightforward way. But I think usually people use media query to check the display size, and then format the content accordingly.

karenfrilya97: @hzxcaryn Okay! Will do something like that.

hzxcaryn: @ifhuang I think it is better to use media query than checking user agent. 
I think checking user agent works for examples like whether clicking a button would lead them to app store, or checking if sharing a link via sms is a valid button. 
However, what Karen wants to achieve here is to alter the layout of the page, depending on the screen width. This `isMobile` here is not literally checking for whether the user is using a mobile or not, but whether the screen width is that of a mobile phone.
I think that we should stick with using media queries for "styling logic", and refactor using the suggestions I wrote in my [review below](https://github.com/googleinterns/gpay-group-buy/pull/141#pullrequestreview-452110650).

Key reason for this is that not all people using their laptops will view the page in full screen. They might split screen (or open developer inspect) and the page would appear pretty narrow. In cases like these we would actually prefer the listings to be wider instead of squashing the listings simply because they are not browsing using a mobile phone.

This is what is happening to the page now if I do that on laptop:
image

ifhuang: Thanks for the detailed comment.
I think your example with split screen on the laptop is a valid use use. In terms of var name, we should probably use sth. like "isSmallScreen" instead of "isMobile".

karenfrilya97: @hzxcaryn That's a nice corner case that supports the use of media query. I agree that media query is more relevant to screen sizes which we are concerned about here. I didn't realise that react-device-detect uses User Agent and wasn't really testing for responsiveness just now, since we are prioritising on the main cases for now. I thought that react-device-detect works for now since it looks cleaner on the code and is immediately understandable, although it doesn't deal with the corner cases.

Since this is related to responsiveness, I will add this particular case to this issue #112 (Make UI responsive).

@ifhuang Okay! That is probably better to avoid this kind of misunderstanding.

PR #146, opened 2020-07-20
Refactor out CommitContext

hzxcaryn:
To be merged after #138 
Refactoring PR, no behavioural changes from #138

# Changes
- After #138, `ListingDetailsContext` seems to be in charge of too many things (it touched on listings, commit, customer). This PR moves out all commit related logic (which includes updating of customer) to `CommitContext`. `ListingDetailsContext` is now just in charge of listing & page related logic.

### DIFF
--- a/web/src/components/customer/listing-details/contexts/CommitContext.tsx
+++ b/web/src/components/customer/listing-details/contexts/CommitContext.tsx
@@ -1,0 +1,140 @@
+/**
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React, {useContext, useState, useEffect} from 'react';
+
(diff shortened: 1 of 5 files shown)


web/src/components/customer/listing-details/contexts/CommitContext.tsx
```diff
@@ -0,0 +1,140 @@
...
+      setCommitStatus(undefined);
+      return;
+    }
+
+    const fetchCommit = async () => {
+      const [commit] = await getCommits({
+        listingId,
+        customerId: customer.id,
+      });
+      if (commit !== undefined) {
```
chshda, line 79: Is it possible for this to be undefined?

hzxcaryn: @chshda Yep, if there are no commits returned from `getCommits` then `commit` will be undefined!
When I do 
```
const [commit] = await getCommits(...);
```
This means that I am doing something equivalent to 
```
const commit = await getCommits(...)[0];
```
So if `getCommits` returns an empty array, `commit` will be undefined

web/src/components/customer/listing-details/contexts/CommitContext.tsx
```diff
@@ -0,0 +1,140 @@
...
+    const commit = await addCommit(
+      {
+        listingId,
+        customerId: customer.id,
+      },
+      idToken
+    );
+    // TODO: Handle addition error
+    setCommitId(commit.id);
+    await refetchCustomer();
```
ifhuang, line 104: do you need to refresh listing?

hzxcaryn: @ifhuang 
Line 105 does `setCommitStatus(commit.commitStatus);` which updates commitStatus.
Listing refreshes itself on commitStatus change in this block of code (highlighted in orange):
image

PR #151, opened 2020-07-22
Add pay for commit endpoint

### LINKED ISSUE #150: Add pay for commit endpoint
Part of #13

hzxcaryn:
Closes #150

# Changes
- Added endpoint to pay for commits
- Added `editAndUpdateRelatedEntity` to datastore layer,  and refactored out `updateEntityInTransaction` in datastore layer

**TODOs, Not in this PR**
- Ensure that commit belongs to the authenticated customer
- Updating of commit interface on client side to include address and contact number
- Validating of paymentData (#153)
- Updating of customer's address and contact number

# Manual Testing
**Non-existent commit id**
Screenshot 2020-07-22 at 1 23 56 PM

**Invalid commit id type**
Screenshot 2020-07-22 at 1 23 45 PM

**Successful payment**
image
**Updated commit after successful payment**
image
**Updated listing after successful payment**
Screenshot 2020-07-22 at 1 11 27 PM

### DIFF
--- a/server/src/interfaces.ts
+++ b/server/src/interfaces.ts
@@ -69,12 +69,20 @@ export interface CommitRequest {
 }
 
 /**
+ * FulfilmentDetails Interface that contains the fields of a fulfilment.
+ */
+interface FulfilmentDetails {
+  name: string;
+  address: string;
+  contactNumber: string; // E164 format
+}
+
+/**
  * CommitPaymentRequest Interface that contains the fields that will be provided
  * by the client in the request body.
  */
 export interface CommitPaymentRequest {
-  deliveryAddress: string;
-  deliveryContactNumber: string; // E164 format
+  fulfilmentDetails: FulfilmentDetails;
 }
 
 /**
--- a/server/src/interfaces.ts
+++ b/server/src/interfaces.ts
@@ -74,7 +74,7 @@ export interface CommitRequest {
  */
 export interface CommitPaymentRequest {
   deliveryAddress: string;
-  deliveryContactNumber: string;
+  deliveryContactNumber: string; // E164 format
 }
 
 /**
--- a/server/src/interfaces.ts
+++ b/server/src/interfaces.ts
@@ -61,7 +61,7 @@ export interface CommitComputedProperties {
 
 /**
  * CommitRequest Interface that contains the fields that will be provided
- * by the client in the POST/PUT request body.
+ * by the client in the request body to create a Commit.
  */
 export interface CommitRequest {
   customerId: number;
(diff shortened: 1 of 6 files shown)


server/src/storage/datastore.ts
```diff
@@ -123,6 +123,32 @@ const updateData = (original: StringKeyObject, updateRule: UpdateRule) => {
...
 
+/**
+ * Updates an entity with the specified kind and id in a transaction
+ * according to the update rules.
+ * @param transaction Transaction that will carry out the update
+ * @param kind Kind of the entity to be updated
+ * @param id id of the entity to be updated
+ * @param updateRules Rules to update the entity
+ */
+const updateEntityInTransaction = async (
```
karenfrilya97, line 134: Good job! This is a great way to refactor operations inside a transaction. Was also thinking of doing this, but hadn't figured out how to do it.

server/src/services/commits.ts
```diff
@@ -113,6 +115,33 @@ const addCommit = async (
   );
 };
 
+/**
+ * Pays for the commit with the specified commitId.
+ * An error will be thrown if the commit is not SUCCESSFUL.
+ * @param commitId Id of the commit to be deleted
+ * @param paymentData Data of the payment request
+ */
+const payForCommit = async (
```
karenfrilya97, line 124: I like how you put `payForCommit` in services and `editCommit` in storage, so storage functions are just simply CRUD and services have functions more specific to the business functions.

server/src/interfaces.ts
```diff
@@ -61,26 +61,42 @@ export interface CommitComputedProperties {
...
   customerId: number;
   listingId: number;
 }
 
+/**
+ * CommitPaymentRequest Interface that contains the fields that will be provided
+ * by the client in the request body.
+ */
+export interface CommitPaymentRequest {
+  deliveryAddress: string;
```
xlai20, line 76: maybe should wrap these kind of information in a container called "FulfilmentDetails", which sounds more like a service waiting to be fulfilled.

server/src/services/commits.ts
```diff
@@ -113,6 +115,33 @@ const addCommit = async (
...
+  commitId: number,
+  paymentData: CommitPaymentRequest
+) => {
+  // Check that listing exists
+  const commit = await commitStorage.getCommit(commitId);
+
+  // Check that listing is successful
+  if (commit.commitStatus !== 'successful') {
```
xlai20, line 132: Did we discuss in previous project meeting that we allow merchants to still sell the product despite not enough customers are committing (and hence not successful)?

hzxcaryn: @xlai20 That's an unsuccessful commit right? I don't think merchants are allowed to sell that, it was the hard requirement we agreed on during meeting last time.

server/src/interfaces.ts
hzxcaryn: @xlai20 
Hmm what do you mean by that? `CommitPaymentRequest` is the type of the payment request body.

Did you mean having a separate `FulfilmentDetails` type? Meaning doing something like:
```
interface FulfilmentDetails {
  deliveryAddress: string;
  deliveryContactNumber: string;
}

export interface CommitPaymentRequest {
  fulfilmentDetails: FulfilmentDetails;
}
```

server/src/interfaces.ts
```diff
@@ -61,26 +61,42 @@ export interface CommitComputedProperties {
...
+  deliveryContactNumber: string; // E164 format
```
chshda, line 77: should we have name also

hzxcaryn: @chshda Ohh that's true, is it in case the receiver of the goods is not the customer themselves?

server/src/interfaces.ts
```diff
@@ -61,26 +61,42 @@ export interface CommitComputedProperties {
...
 export interface CommitRequest {
   customerId: number;
   listingId: number;
 }
 
+/**
+ * CommitPaymentRequest Interface that contains the fields that will be provided
+ * by the client in the request body.
+ */
+export interface CommitPaymentRequest {
```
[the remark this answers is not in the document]
hzxcaryn: @ifhuang Oh I was thinking we shouldnt use that because they might not want to send items that they purchase to the same address everytime. 
In the future, #167 will be implemented and the default delivery details will be auto-filled for the user on the frontend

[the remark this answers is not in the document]
ifhuang: I was thinking it's annoying that customer need to input address every time. But your #167 (set default and auto-fill) should solve that.

server/src/interfaces.ts
xlai20: yes, that was what I meant.

PR #152, opened 2020-07-22
Add View Listing Details page for Merchant

karenfrilya97:
### Changes
**[Main Change]**
- **Add merchant's listing-details page.**
  - This page uses ListingDescription component in components/common.

**[Other Changes]**
- **Add priceFontSize and commitProgressFontSize props to ListingDescription.**
  - So that the one shown in merchant's listing-details page can show larger commit progress font and smaller price font.
  - Commit progress is more important/helpful to merchant than price, unlike to customer.
- **Change commitProgress largerFont prop to fontSize.**
  - There are now 3 different font size options instead of 2.
- **Change redirection path in the buttons in MerchantSideBar to absolute path.**
  - Listing-details page has an extra nested layer ('listing/:id' compared to other pages like 'home') which will cause the relative path to lead to the wrong absolute url.

### Test
image

### Linked Issues
Part of #5, #7, #8, #9.

### To-do
- Prevent merchant from viewing details of another merchant's listing.


web/src/components/common/CommitProgress.tsx
```diff
@@ -42,8 +43,16 @@ const CommitProgressContainer = styled.div`
...
   text-align: right;
-  font-size: ${({largerFont}: CommitProgressContainerProps) =>
-    largerFont ? '1em' : '0.8em'};
+  font-size: ${({fontSize}: CommitProgressContainerProps) => {
+    switch (fontSize) {
+      case 'large':
+        return '2em';
+      case 'medium':
+        return '1em';
+      default:
```
hzxcaryn, line 52: nit: Maybe we can do this so that it's clearer that 'small' goes under the default case too
```suggestion
      case 'small':
      default:
```

web/src/components/common/ListingDescription.tsx
```diff
@@ -61,17 +62,23 @@ const ListingDescription: React.FC<ListingDescriptionProps> = ({
...
+  commitProgressFontSize,
 }) => (
   <Container>
     <DeadlineTag deadline={deadline} largerFont />
     <ListingName>{name}</ListingName>
-    <ListingPrice price={price} oldPrice={oldPrice} largerFont />
+    <ListingPrice
+      price={price}
+      oldPrice={oldPrice}
+      largerFont={priceFontSize !== 'small'}
```
hzxcaryn, line 74: I think this part is really weird, like allowing 3 options for `priceFontSize` but it really only resolves to 2 values ('large' and 'medium' has no difference).

web/src/components/common/CommitProgress.tsx
karenfrilya97: @hzxcaryn Okay!

web/src/components/common/ListingDescription.tsx
karenfrilya97: @hzxcaryn Okay! As discussed over chat, will only allow `priceFontSize` to accept `'medium'` or `'large'`.

web/src/components/common/ListingDescription.tsx
```diff
@@ -61,17 +62,23 @@ const ListingDescription: React.FC<ListingDescriptionProps> = ({
...
+  commitProgressFontSize,
 }) => (
   <Container>
     <DeadlineTag deadline={deadline} largerFont />
     <ListingName>{name}</ListingName>
-    <ListingPrice price={price} oldPrice={oldPrice} largerFont />
+    <ListingPrice
+      price={price}
+      oldPrice={oldPrice}
+      largerFont={priceFontSize === 'large'}
```
xlai20, line 74: Can we do the same for ListingPrice's font size as what we do with the CommitProgress? That is, instead of a boolean variable of "largerFont", just use a FontSize enum (can share the same enum or create its own size enum).

web/src/components/merchant/listing-details/ListingDetails.tsx
```diff
@@ -0,0 +1,71 @@
...
+  background-color: var(--light-gray);
+`;
+
+interface ListingDetailsProps {
+  listingId: number;
+}
+
+/**
+ * ListingDetails that contains full listing details. It is different from
+ * ListingDetails in components/customer/ because it doesn't contain merchant details
```
xlai20, line 41: What does the "it" refer to in this sentence? Please disambiguate.

web/src/components/merchant/listing-details/ListingDetails.tsx
```diff
@@ -0,0 +1,71 @@
...
+  object-fit: cover;
+  background-color: var(--light-gray);
+`;
+
+interface ListingDetailsProps {
+  listingId: number;
+}
+
+/**
+ * ListingDetails that contains full listing details. It is different from
```
xlai20, line 40: change to "...contains details of a listing with specified listingId."

web/src/components/common/ListingDescription.tsx
karenfrilya97: @xlai20 Okay!

web/src/components/merchant/listing-details/ListingDetails.tsx
karenfrilya97: @xlai20 Okay! I change to 'the former'.

web/src/components/merchant/listing-details/ListingDetails.tsx
karenfrilya97: @xlai20 Okay!

web/src/interfaces.ts
```diff
@@ -24,6 +24,29 @@ export interface Image {
   alt?: string;
 }
 
+/**
+ * FontSize type that provides several preset options for a font size.
+ */
+type FontSize = 'small' | 'medium' | 'large';
+
+/**
+ * CommitProgressFontSize type that provides several preset options for the font size of CommitProgress.
```
ifhuang, line 33: it exceeds 80 chars. can you break this line into two lines?

web/src/components/merchant/listing-details/ListingDetails.tsx
```diff
@@ -0,0 +1,71 @@
...
+  object-fit: cover;
+  background-color: var(--light-gray);
+`;
+
+interface ListingDetailsProps {
+  listingId: number;
+}
+
+/**
+ * ListingDetails that contains details of a listing with specified listingId. This component is different from
```
ifhuang, line 40: same here. pls break into two lines

web/src/interfaces.ts
karenfrilya97: @ifhuang Okay!

web/src/components/merchant/listing-details/ListingDetails.tsx
karenfrilya97: @ifhuang Okay!

PR #158, opened 2020-07-23
Add customer-facing merchant profile page

### LINKED ISSUE #120: Add customer-facing merchant profile/listings page
When customers click on the merchant profile section in the listings details page, they will be linked to `/merchant/:id`  [(UI Mock)](https://www.figma.com/file/x0PWXRWxwSopiixYv41mK7/GPay-GroupBuy?node-id=1%3A40602)

hzxcaryn:
Closes #120 

# Changes
- Added customer-facing merchant profile page

**Screenshots of some existing customers**
Screenshot 2020-07-23 at 5 32 49 PM

Screenshot 2020-07-23 at 2 16 46 PM


web/src/interfaces.ts
```diff
@@ -84,6 +84,12 @@ export interface Listing extends ListingPayload {
   listingStatus: ListingStatus;
 }
 
+/**
+ * ListingQuery Interface that contains the fields of the query that
+ * would be sent to the server to query for commits.
```
ifhuang, line 89: query for listings

web/src/Routes.tsx
```diff
@@ -28,7 +30,7 @@ const Routes: React.FC = () => (
   <Suspense fallback={<Loading />}>
     <Switch>
       <Route exact path="/design-samples" component={DesignSamplesPage} />
-      <Route path="/merchant/" component={MerchantRoutes} />
+      <Route path={MERCHANT_ROOT} component={MerchantRoutes} />
```
xlai20, line 33: Some routes have literal paths; some routes have paths imported from constants. What's the rationale behind this distinction? Can we have some consistency in this routes file?

web/src/api.ts
```diff
@@ -139,9 +140,9 @@ export const addListing = async (
  * Fetches all Listings.
  * @param queryParams Query parameters used to filter listings to retrieve.
```
xlai20, line 141: Update the comment when API signature is changed.

web/src/interfaces.ts
hzxcaryn: @ifhuang Oops it slipped my mind to change it! Will change!

web/src/Routes.tsx
hzxcaryn: @xlai20 
Because this commit https://github.com/googleinterns/gpay-group-buy/pull/158/commits/e31924f4c47d623aef0fe8925717002fb601ec47 adds `MERCHANT_ROOT` for `ListingCollection` so that we don't have to directly pass in `/merchant/` to `listingRootPath` props, as I did not really want components don't have to be aware of what merchant root is. 
And since I added `MERCHANT_ROOT` to constants, I felt that the `Routes.tsx` should use it too to be consistent. (I didn't want the case where someone changes the path in `Route` and make `MERCHANT_ROOT` outdated.
At the same time I am reluctant to change the rest of the routes to be imported because they will only be used once in this file and mainly because it would be less clear how paths are routed in the `Routes.tsx` file.
For example, if we have a constant `const CUSTOMER_ROOT = '/'`, and we do
```
      <Route path={MERCHANT_ROOT} component={MerchantRoutes} />
      <Route path={CUSTOMER_ROOT} component={CustomerRoutes} />
```
Then it becomes unclear if path '/' is used or not (which is kind of important in routing), and also, it looks like the order can be changed up like so:
```
      <Route path={CUSTOMER_ROOT} component={CustomerRoutes} />
      <Route path={MERCHANT_ROOT} component={MerchantRoutes} />
```
Which should not be allowed because `path='/'` should be the last path, else every single route will be routed to `CustomerRoutes`.
Hmm what do you think I should do to make this better?

web/src/api.ts
hzxcaryn: @xlai20 It's the same though :o I only made changed the type to be more specific rather than accepting any filter string fields.

PR #159, opened 2020-07-23
Configure composite indexes for Datastore using index.yaml file

karenfrilya97:
### Changes
- Added index.yaml file containing required composite indexes.
- Added instructions for configuring composite indexes on README.md.


server/README.md
```diff
@@ -108,6 +108,22 @@ Each of these folders contain the following files:
 Each file in this folder contains a group of constants that are closely related.
 The file name specifies the constants inside it.
 
+## Configuring Database
+
+### Configuring Indexes
+
+In Datastore, single-property indexes are built in and do not need to be created.
+
+To make changes to composite indexes, edit `index.yaml` file accordingly and then deploy using the following command:
```
hzxcaryn, line 117: nit: can we have a newline here hehe
```suggestion
To make changes to composite indexes, edit `index.yaml` file accordingly and then deploy using the following command:

```

server/index.yaml
```diff
@@ -0,0 +1,20 @@
...
+  properties:
+  - name: listingId
+  - name: commitStatus
+  - name: createdAt
+
+- kind: Listing
+  properties:
+  - name: merchantId
+  - name: listingStatus
+  - name: deadline
```
ifhuang, line 20: do you need 'direction: desc' for deadline?

karenfrilya97: @ifhuang Hmm I think I prefer the default direction, which is 'asc' so that the one with the nearest deadline will appear first for ongoing listings. And for successful listings that require the merchant to deliver the items, I also want the oldest one to appear first, so that the merchant will send these items first.
I guess for completed and unsuccessful listings, it might be good to have 'direction: desc' but I think that's not as important and can be added later on when needed.

server/README.md
karenfrilya97: @hzxcaryn okay!

server/index.yaml
karenfrilya97: @ifhuang I prefer the default direction which is ascending. For 'ongoing' listings, I would like to display the ones with the nearest deadline first, as that seems to be the most urgent. For 'successful' listings, I would like to display the oldest ones first, so that the merchant will send the items for these listings first (so that the customers don't have to wait for too long).

PR #160, opened 2020-07-23
Sort listings by deadline

karenfrilya97:
#### Merchant listings are sorted by deadline in ascending order:
image

#### All listings are sorted by deadline in ascending order:
image


server/src/interfaces.ts
```diff
@@ -177,6 +177,14 @@ export interface Filter {
   value: any;
 }
 
+/**
+ * OrderRule Interface that contains order rules for sorting.
```
ifhuang, line 181: add comment about the order if descending not specified

karenfrilya97: @ifhuang okay!

karenfrilya97: @ifhuang Added the comment in datastore.ts in the JSDoc description for `getAll` function [here](https://github.com/googleinterns/gpay-group-buy/pull/160/commits/6f8cdbac1fd29909eb449142563514778c9cb2e3). This is because the default direction is only specified by Datastore's `query.order()` function. But this interface here doesn't specify the default direction.

PR #162, opened 2020-07-24
Refactor datastore to use HOF transaction

### LINKED ISSUE #161: Refactor datastore transactions
There are multiple use cases of datastore transactions in our app, as we want these operations to be atomic.
Examples:  Updating the `numCommits` in `Listing` when a new `Commit` entity is created, updating `numCommits` in `Listing` when a `Commit` entity is deleted, updating `numPaid` in `Listing` when a `Commit` has its `commitStatus` changed to 'paid', etc.

This results in many chunky methods in our `storage/datastore` layer, such as `addAndUpdateRelatedEntity` and `deleteAndUpdateRelatedEntity`, which all creates a transaction and does the adding/deleting and updating in the same block of code inside the same transaction. 

This becomes a problem when we need more and more of such transactions -- we would need to create more of `doThisAndDoThisAndDoThat` functions in datastore. This is not scalable and makes the codebase hard to maintain and messy in the longer term.

My idea is to create a higher order function that wraps all operations that should be completed in the same transaction. 
Something like:
```
makeTransaction(
  transactionOp1.
  transactionOp2.
  ...,
  transactionOpN.
);
```

hzxcaryn:
Closes #161
**Please refer to #161 for background for this PR**

# Changes
- Added higher order function `makeTransaction` in datastore layer
- Added "building blocks" for transactions, such as `getEntityInTransaction` and other functions ending with `InTransaction` 
- Replaced all existing transaction functions like `addAndUpdateRelatedEntity` with `makeTransaction` and the appropriate "building blocks". 

So now, instead of
```
  const commitId = await addAndUpdateRelatedEntity(
    COMMIT_KIND,
    commit,
    LISTING_KIND,
    commit.listingId,
    [
      {
        property: 'numCommits',
        op: 'add',
        value: 1,
      },
    ],
    uniqueProperties
  );
```
We now do
```
  const [commitId] = await makeTransaction(
    addInTransaction(COMMIT_KIND, commit, uniqueProperties),
    updateInTransaction(LISTING_KIND, commit.listingId, [
      {
        property: 'numCommits',
        op: 'add',
        value: 1,
      },
    ])
  );
```

# Testing
**Successful GET for listings (shows that non-transaction operations works)**
Screenshot 2020-07-28 at 7 53 44 PM

**Successful DELETE for commits, and the updating of `numCommits` in Listing (shows that transaction works)**
image
image

**Successful POST for commits, and the updating of `numCommits` in Listing (shows that transaction works)**
Screenshot 2020-07-28 at 8 54 33 PM
image

**Failed POST for commits (shows that failing in transaction works)**
Screenshot 2020-07-28 at 8 03 19 PM

### DIFF
--- a/server/src/storage/datastore.ts
+++ b/server/src/storage/datastore.ts
@@ -77,11 +77,10 @@ const extractAndAppendId = (res: Entity) => {
 
 /**
  * A Datastore wrapper that gets a particular entity with the specified Kind and id.
- * If actor is not specified, gets using datastore.
+ * If transaction is not specified, gets using datastore.
  * @param kind The Kind that is being retrieved
  * @param id The id of the Entity being retrieved
- * @param transaction Transaction that will carry out the retrieval,
- * if not specified, datastore will carry out the operation
+ * @param transaction Transaction that will carry out the retrieval
  */
 export const getEntity = async (
   kind: string,
@@ -113,14 +112,13 @@ export const getEntityInTransaction = async (kind: string, id: number) => (
 
 /**
  * A Datastore wrapper that gets all entities of a specified Kind.
- * If actor is not specified, gets using datastore.
+ * If transaction is not specified, gets using datastore.
  * @param kind The Kind that is being queried
  * @param filters Any filters that will be applied to the query
  * @param orderRules Any order rules that will be used to sort the query result.
  * If an orderRule doesn't have a descending property specified, the default
  * direction is ascending.
- * @param transaction Transaction that will carry out the query,
- * if not specified, datastore will carry out the operation
+ * @param transaction Transaction that will carry out the query
  */
 export const getAllEntities = async (
   kind: string,
@@ -215,12 +213,11 @@ const insertUniqueEntity = async (
  * a function that resolves to the id of the inserted entity.
  * If uniqueProperties are specified, the entity would not be added if another entity with the
  * same value for the unique properties already exists.
- * If actor is not specified, adds using datastore.
+ * If transaction is not specified, adds using datastore.
  * @param kind The Kind of the Entity to be added
  * @param data The data of the Entity to be added
  * @param uniqueProperties The properties that should be unique for the specified kind
- * @param transaction Transaction that will carry out the adding,
- * if not specified, datastore will carry out the operation
+ * @param transaction Transaction that will carry out the adding
  */
 const addEntityHelper = async (
   kind: string,
@@ -252,7 +249,6 @@ const addEntityHelper = async (
  * the id of the inserted entity.
  * If uniqueProperties are specified, the entity would not be added if another entity with the
  * same value for the unique properties already exists.
- * If actor is not specified, adds using datastore.
  * @param kind The Kind of the Entity to be added
  * @param data The data of the Entity to be added
  * @param uniqueProperties The properties that should be unique for the specified kind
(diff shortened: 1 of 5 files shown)


server/src/storage/datastore.ts
```diff
@@ -22,6 +22,40 @@ import {UpdateRule} from './interfaces';
...
+type TransactionResponse = number | void;
+type TransactionOperation = (
+  transaction: Transaction
+) => Promise<TransactionResponse>;
+
+/**
+ * A higher order function that creates a transaction and carries out all
+ * input datastore operations in the same transaction.
+ * Returns the response of each operation carried out in the transaction in
+ * an array, in the same order they are called.
```
ifhuang, line 34: Can you also add comment about what happen if some operation failed?

server/src/storage/datastore.ts
```diff
@@ -22,6 +22,40 @@ import {UpdateRule} from './interfaces';
...
+  ...opFns: TransactionOperation[]
+): Promise<TransactionResponse[]> => {
+  const transaction = datastore.transaction();
+
+  const res = [];
+
+  try {
+    await transaction.run();
+
+    for (const fn of opFns) {
```
ifhuang, line 47: consider renaming fn and opFns, they're not very straightforward.
e.g. operation and operations

server/src/storage/datastore.ts
```diff
@@ -38,49 +72,90 @@ const extractAndAppendId = (res: Entity) => {
 
 /**
  * A Datastore wrapper that gets a particular entity with the specified Kind and id.
- * @param kind The Kind that is being queried
- * @param id The id of the Entity being queried
+ * If actor is not specified, gets using datastore.
+ * @param kind The Kind that is being retrieved
+ * @param id The id of the Entity being retrieved
+ * @param actor Datastore or Transaction that will carry out the retrieval
```
ifhuang, line 78: What's the difference between passing in Datastore and passing in Transaction? Also can we have a more specific name than actor

hzxcaryn: @ifhuang You pass in a transaction object if you want that particular transaction to carry out the operation. The same transaction object has to be passed in for all operations that needs to be atomic. 
The datastore object (the default when nothing is passed in) just carries out the datastore operation (not in any transaction).

hzxcaryn: @ifhuang Okay, so basically I'll be moving the actor code into the implementation? So I have this line on the first line of each function?
```
const actor = transaction || datastore;
```

server/src/storage/datastore.ts
```diff
@@ -108,6 +196,88 @@ const uniqueInsertInTransaction = async (
...
+
+/**
+ * A higher order function.
+ * Returns a function that takes in a transaction object and adds
+ * the entity in the transaction.
+ * @param kind The Kind of the Entity to be added
+ * @param data The data of the Entity to be added
+ * @param uniqueProperties The properties that should be unique for the specified kind
+ */
+export const addInTransaction = (
```
chshda, line 274: might be better to rename as addEntityInTransaction to be consistent

server/src/storage/datastore.ts
```diff
@@ -38,60 +77,109 @@ const extractAndAppendId = (res: Entity) => {
 
 /**
  * A Datastore wrapper that gets a particular entity with the specified Kind and id.
- * @param kind The Kind that is being queried
- * @param id The id of the Entity being queried
+ * If actor is not specified, gets using datastore.
```
ifhuang, line 80: update 'actor' to 'transaction' in comments

hzxcaryn: @ifhuang Oh I missed that our when editing! I think I will just remove this line since it is kind of repetition for
```
 * @param transaction Transaction that will carry out the retrieval,
 * if not specified, datastore will carry out the operation
```

server/src/storage/datastore.ts
hzxcaryn: @chshda Actually all the functions ending with `InTransaction` don't have the word `Entity`. I will add it for all of them then.

PR #171, opened 2020-07-27
Add delivery details form upon payment

hzxcaryn:
Closes #154, Closes #15 

# Changes 
- Added delivery details form prompt that appears upon clicking "Pay"
- Basic validations for form (w/o error messages as of now)

**TODO, not in this PR:** 
- Refactor `MobileFormRow` to combine with `FormRow` (right now, because of the difference in styles, I made a separate `MobileFormRow` so that the changes for this PR would not be too big
- Add phone number validations for phone number in form

## Screenshots
**Before paying**
Screenshot 2020-07-27 at 8 07 36 PM

**After clicking pay (confirm button is disabled)**
Screenshot 2020-07-27 at 8 43 25 PM

**After filling in details (confirm button is enabled)**
Screenshot 2020-07-27 at 8 43 10 PM

**After clicking confirm**
Screenshot 2020-07-27 at 7 50 44 PM
Screenshot 2020-07-27 at 7 50 54 PM

PR #173, opened 2020-07-28
Refactor Form Row and Mobile Form Row

hzxcaryn:
To be merged after #171

# Changes
-  #171 added a `MobileFormRow` which is very similar to `FormRow` except for the styles. This PR refactors `MobileFormRow` into `FormRow`.

**Screenshots to show that the fields still look like how they used to**
Screenshot 2020-07-28 at 2 27 13 PM
Screenshot 2020-07-28 at 2 25 47 PM
image


web/src/components/common/FormRow.tsx
```diff
@@ -36,7 +54,11 @@ const StyledCol = styled(Col)`
   align-items: left;
 `;
 
-const Label = styled.label`
+interface LabelProps {
+  stacked?: boolean;
+}
+
+const nonStackedLabelStyle = css`
```
karenfrilya97, line 61: Is there a name we can use that doesn't use a negative? Hahaha Something like inlineLabel?

hzxcaryn: @karenfrilya97 HAHAHA naming is hard! Sounds good, thanks!

web/src/components/common/FormRow.tsx
```diff
@@ -81,28 +87,105 @@ const ErrorContainer = styled.div`
...
+  inputStyle?: FormInputStyle;
 }
 
 /**
- * This is a row in a form, consisting of a label and an input field shown
- * side by side. This also contains a container for error message which is
+ * FormRow is a row in a form, consisting of a label and an input field, styled according
+ * to inputStyle. FormRow also contains a container for error message which is
  * displayed below the input field where applicable.
+ * Index is the index of the row in the form.
```
ifhuang, line 152: pls use @param

hzxcaryn: @ifhuang At first I was reluctant to use `@param` because they are props so `@param index <desc>` is does not really make sense.
But I guess I can do  `@param props.index <desc>`

PR #174, opened 2020-07-28
Add query by ids query param to /listings endpoint

hzxcaryn:
# Changes
- Added `/listings?ids=:ids` endpoint, which fetches all specified listing ids (will be required by My Commits page)
  * Takes in ids in the format `/listings?ids=<id1>,<id2>,<id3>,...`
  * Consumers of endpoint are not allowed to filter by other query params if `ids` query is specified
- Added tests 

# Manual testing
**When trying to query by other query params white `ids` is specified**
Screenshot 2020-07-28 at 6 42 18 PM

**When trying to query with invalid listing id**
Screenshot 2020-07-28 at 6 41 49 PM

**Successful query**
Screenshot 2020-07-28 at 6 39 04 PM


server/src/storage/datastore.ts
```diff
@@ -51,6 +51,24 @@ export const get = async (kind: string, id: number) => {
   }
 };
 
+/**
+ * A Datastore wrapper that gets a particular entity with the specified Kind and id.
+ * @param kind The Kind that is being queried
+ * @param id The id of the Entity being queried
```
ifhuang, line 57: nit: ids

server/src/api/listings.ts
```diff
@@ -25,19 +25,42 @@ import {listingService} from '../services';
 
 const listingRouter = Router();
 
+/**
+ * Handles the get request for retrieving all listings that satisfy the query
+ * parameters.
+ * Querying by ids is a restricted query, once the id field is specified, requester
```
ifhuang, line 31: this is impl details, and we'd better to shift it to line#39

server/src/api/listings.ts
```diff
@@ -25,19 +25,42 @@ import {listingService} from '../services';
...
   async (req: Request, res: Response, next: NextFunction) => {
     try {
+      const {ids: idsStr} = req.query;
+      if (idsStr !== undefined) {
+        if (Object.keys(req.query).length > 1) {
+          res.sendStatus(400);
+          return;
+        }
+
+        const ids = (idsStr as string).split(',').map(idStr => {
```
ifhuang, line 46: should we allow empty ids?

server/src/api/listings.ts
```diff
@@ -25,19 +25,42 @@ import {listingService} from '../services';
...
+        });
+        const listings = await listingService.getAllListingsWithIds(ids);
+        res.json(listings);
+        return;
+      }
+
       const filters = Object.keys(req.query).map(key => ({
         property: key,
-        value:
-          key === 'customerId' || key === 'merchantId'
```
ifhuang, line 35: you removed customerId. is it expected?

server/tests/integration/listings.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    test('Should fetch all listings', async () => {
+      const expectedListingData = listingsFixtures.responseData;
+
+      const res = await request(app).get('/listings');
+
+      expect(res.status).toBe(200);
+      expect(res.body).toMatchObject(expectedListingData);
+    });
+
+    test('Should filter by listing ids', async () => {
```
ifhuang, line 34: we should split this into two tests: one with single id, another one with multiple ids.
we should try to make tests simple, and avoid extra logic (e.g. map/forEach here).

server/src/api/listings.ts
hzxcaryn: @ifhuang Yep it's not needed anymore because initially I wanted to query Listings by `customerid`, but since `customerId` is not a field it makes the handling of this query quite messy. So now api consumers should first query for Commits with `customerId` and then query for Listings using the `listingId`s from the Commits.

server/src/api/listings.ts
hzxcaryn: @ifhuang Good question, I think I would make it a 400.

server/tests/integration/listings.test.ts
```diff
@@ -0,0 +1,85 @@
...
+
+      const res = await request(app).get('/listings').query({
+        ids: targetListingId,
+        merchantId: targetMerchantId,
+      });
+
+      expect(res.status).toBe(400);
+    });
+
+    test('Should not filter by empty listing ids', async () => {
```
chshda, line 77: will be good to add a case for invalid listing id as well

PR #176, opened 2020-07-29
Add listing card with commit status

hzxcaryn:
Merge after #174 

# Changes
- Added `ListingCardWithCommitStatus` which displays the customer's commit status of the listing. [(UI mock](https://www.figma.com/file/x0PWXRWxwSopiixYv41mK7/GPay-GroupBuy?node-id=1%3A40239))
- Added some samples to the design-samples page

**TODO for future PRs**
- Fix Card height to follow height of content instead of height of image (see screenshots to see what I mean)
  * Not too sure when this problem happened but it happened before this PR, so I would consider it out of scope for this PR

# Screenshots
**This is how it looks like now**
Screenshot 2020-07-29 at 4 48 19 PM

**This is how it should look like after I fix the Card height bug**
Screenshot 2020-07-29 at 4 48 02 PM

# How to test
Fetch my branch and 
```
cd web
yarn
yarn start
```
And navigate to `http://localhost:3000/design-samples`.

### DIFF
--- a/web/src/components/customer/my-commits/CommitStatusBadge.tsx
+++ b/web/src/components/customer/my-commits/CommitStatusBadge.tsx
@@ -1,0 +1,96 @@
+/**
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React from 'react';
+
+import {CommitStatus} from 'interfaces';
+import styled, {css} from 'styled-components';
+
+const dangerStyle = css`
+  background-color: var(--bright-red);
+  color: white;
+`;
+
+const successStyle = css`
+  background-color: var(--pale-green);
+  color: var(--green);
+`;
+
+const grayedStyle = css`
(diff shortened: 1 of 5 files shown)


web/src/components/customer/my-commits/CommitStatusBadge.tsx
```diff
@@ -0,0 +1,97 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React from 'react';
+
+import {CommitStatus} from 'interfaces';
+import styled, {css} from 'styled-components';
+
+const dangerStyle = css`
```
xlai20, line 22: Maybe "actionStyle" is a better name?

web/src/components/customer/my-commits/CommitStatusBadge.tsx
```diff
@@ -0,0 +1,97 @@
...
+      default:
+        return '';
+    }
+  }}/* stylelint-enable value-keyword-case */
+`;
+
+const getBadgeText = (commitStatus: CommitStatus) => {
+  switch (commitStatus) {
+    case 'successful':
+      return 'Pay';
```
xlai20, line 69: Honestly, these "badges" look really like clickable buttons instead of text-based status.

Suggestion:
1. Change the wording "Pay" to a word that doesn't sound like an action, similar to how you name other status like "waiting delivery", "failed" etc.
2. Take away the rounded border to make it look less like a button (or whatever change that make it look more like a badge instead of a button).

hzxcaryn: @xlai20 Actually I think only "PAY" looks like a button and actually it's intended! I was thinking of leading the customer to pay directly if they click on the listing with the "PAY" badges. (not implemented yet)

web/src/components/customer/my-commits/CommitStatusBadge.tsx
hzxcaryn: @xlai20 Okay will rename it to `pendingActionStyle` :D

PR #187, opened 2020-08-03
Store whether customer is existing customer in localstorage

### LINKED ISSUE #186: Automatically log in existing customers when they revisit the Spot
# Background
Currently, we do not call the getIdentity API for customer the moment they enter our app. This is to improve UX of first-time visitors as they do not have to "sign in" before checking out our app (as specified by [Spot UX docs](https://developers.google.com/pay/spot/eap/guides/customizable-experience#timely_consents)). Instead, we will call the getIdentity API when they decide to do actions that require them to sign in, such as committing to a listing, or trying to visit the My Commits page.

# Problem
While this is good for first time users, it becomes annoying for existing customers to not be automatically signed in whenever they visit the Spot. 

# Proposed solution
We can improve UX for existing customers by automatically signing in only customers who have visited our Spot before, and maintain the previous behaviour for first-time visitors.
To do so, we can store a flag in localStorage to track if the customer is an existing customer or not. If it is an existing customer, we will automatically call the getIdentity API.

hzxcaryn:
Closes #186
**Please read #186 for background of this PR**

# Changes
- Added `isExistingCustomer` flag to localstorage so that we can identity if it is an existing customer
  * If existing customer, CustomerContext will log in customer when they enter the app
  * else, do nothing, retain original behaviour
- Added `useLocalStorageHook` that handles setting and getting of state to persist in localstorage
- Removed "Get Identity" button on Explore page because we don't need it anymore :) 

This is how localstorage looks like after I tried logging in once (retains after page refreshes)
Screenshot 2020-08-03 at 6 43 22 PM

### DIFF
--- a/web/src/components/common/hooks/useLocalStorage.ts
+++ b/web/src/components/common/hooks/useLocalStorage.ts
@@ -1,0 +1,62 @@
+/**
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {useState, useCallback} from 'react';
+
+/**
+ * Code referenced from: https://usehooks.com/useLocalStorage/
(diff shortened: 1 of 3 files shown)


hzxcaryn:
@ifhuang 
> Instead of a boolean, did you consider to store existing customer info instead?

Yep thought of that in the past when I was deciding if I should store customer info in localstorage or in a Context. Currently we are storing our customer data in `CustomerContext`.

I did not want to store info in the localstorage because it persists across sessions and I do not want the data I store to be outdated. 
- Customer could have revoked permissions for our microapp to access their data since their last visit (if that's the case, it would probably not make sense for the user to see thier data still displayed on the app)
- Customer might have signed out of a google account to use another google account. If we store customer info in localstorage, we would be displaying the wrong customer's info. The only way to find out if the user has changed is to call the `getIdentity` API. If that's the case, there is no point storing customer info in localstorage and storing a boolean will suffice

Issue #167, opened 2020-07-24
Allow customers to set default delivery details 


hzxcaryn:
"Use as default" button from https://www.figma.com/file/x0PWXRWxwSopiixYv41mK7/GPay-GroupBuy?node-id=0%3A1 that saves these delivery details to the customer Entity.
The next time user makes a payment, this default delivery details will be auto-filled for the user

PR #195, opened 2020-08-04
Add endpoint for updating customer info

hzxcaryn:
Part of #167 

# Changes
- Added PATCH endpoint for `/customers/:customerId` for updating of `defaultFulfilmentDetails`. This will be useful when we want to save customer info for future reference after they have filled in the delivery details form once.

# Testing efforts
**Non-existent customer**
image

**Invalid customer id**
image

**Invalid phone number**
Screenshot 2020-08-05 at 7 59 13 PM

**Successful update**
Screenshot 2020-08-05 at 8 03 20 PM


server/src/services/customers.ts
```diff
@@ -45,4 +49,38 @@ const addCustomer = async (
...
+ * or if the customer does not exist.
+ * @param customerId Id of the customer to update
+ * @param fieldsToUpdate Fields of the customer to be updated
+ */
+const updateCustomer = async (
+  customerId: number,
+  fieldsToUpdate: Partial<CustomerPayload>
+) => {
+  // Check that customer exists
+  await customerStorage.getCustomer(customerId);
```
karenfrilya97, line 64: Is this query necessary? Won't there already be an error thrown during update if a customer with given customerId doesn't exist?

server/src/api/customers.ts
```diff
@@ -83,4 +84,38 @@ customerRouter.post(
...
+    try {
+      if (Number.isNaN(customerId)) {
+        throw new Error('Invalid customerId.');
+      }
+
+      const modifiedCustomer = await customerService.updateCustomer(
+        customerId,
+        customerData
+      );
+      res.status(200).send(modifiedCustomer);
```
karenfrilya97, line 113: I think there should be 'Content-Location' header here as well, since we are returning an object that can be retrieved using another method.

server/src/api/customers.ts
```diff
@@ -83,4 +84,38 @@ customerRouter.post(
...
+  customerAuth,
+  validateAndFormatPhoneNumber(
+    (body: Partial<CustomerPayload>) => body.contactNumber,
+    (body: Partial<CustomerPayload>, phoneNumber: string) =>
+      (body.contactNumber = phoneNumber)
+  ),
+  async (req: Request, res: Response, next: NextFunction) => {
+    const {customerId: customerIdStr} = req.params;
+    const customerId = Number(customerIdStr);
+    const customerData = req.body;
```
karenfrilya97, line 102: Do you think `customerData` may sound like it contains all the fields of a Customer and not just those to be updated?

server/src/api/customers.ts
hzxcaryn: @karenfrilya97 
According to https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html section **14.14 Content-Location** it doesn't seem necessary especially since it is the exact same url that the user has requested for?

server/src/services/customers.ts
hzxcaryn: @karenfrilya97 
Actually right now datastore `updateEntity` does not throw an appropriate error if customer doesn't exist.
Indeed some kind of error will be thrown when updating a customer with invalid customerId because in `updateEntity` we have:
 ```
  const key = datastore.key([kind, id]);
  const [data] = await datastore.get(key);
```
And `data` will be `undefined` if the `id` in the key does not exist. So right now our `updateEntity` will throw errors because then it would be working with something that's undefined.
if you think it's better to not do the check in services layer then I guess I can modify `updateEntity` to throw an appropriate error when `data === undefined`!

File: server/src/api/customers.ts (hunk 3 above)
Comment (hzxcaryn) [reply server/src/api/customers.ts:102]: @karenfrilya97 Okay will change that!

File: server/src/services/customers.ts (hunk 1 above)
Comment (karenfrilya97) [reply server/src/services/customers.ts]: @hzxcaryn Personally, I prefer to throw the appropriate error in `updateEntity` function. 

I think that updating something that doesn't exist should not be allowed. This would also reduce repetition in the services layer.

Also, even though get function is relatively fast and has O(1) complexity, I prefer not to make an extra trip to the database unless it's important.

File: server/src/api/customers.ts (hunk 2 above)
Comment (karenfrilya97) [reply server/src/api/customers.ts:115]: @hzxcaryn Oh okay! Thanks for clarifying.

File: server/src/services/customers.ts (hunk 1 above)
Comment (hzxcaryn) [reply server/src/services/customers.ts]: @karenfrilya97 Oh I already changed `updateEntity` to call `getEntity` when I resolved the comment a few days ago, forgot to mention 😓 This means that `updateEntity` would throw the same error as `getIdentity` if entity does not exist.

File: server/src/storage/datastore.ts [hunk 4]
```diff
@@ -80,11 +80,13 @@ const extractAndAppendId = (res: Entity): Entity => {
  * If transaction is not specified, gets using datastore.
  * @param kind The Kind that is being retrieved
  * @param id The id of the Entity being retrieved
+ * @param appendId A flag to indicate if id should be appended to Entity
  * @param transaction Transaction that will carry out the retrieval
  */
 export const getEntity = async (
   kind: string,
   id: number,
+  appendId = true,
```
xlai20, line 89: probably rename to "shouldAppendId"

server/src/storage/datastore.ts
```diff
@@ -105,20 +107,25 @@ export const getEntity = async (
...
  * @param id The id of the Entity being retrieved
+ * @param appendId A flag to indicate if id should be appended to Entity
  */
-export const getEntityInTransaction = async (kind: string, id: number) => (
-  transaction: Transaction
-) => getEntity(kind, id, transaction);
+export const getEntityInTransaction = async (
+  kind: string,
+  id: number,
+  appendId = true
```
xlai20, line 115: ditto

Issue #10, opened 2020-06-11
Merchant can indicate item sent


karenfrilya97:
https://www.figma.com/file/x0PWXRWxwSopiixYv41mK7/GPay-GroupBuy?node-id=44%3A612

PR #197, opened 2020-08-04
Create Complete Commit API

karenfrilya97:
### Changes
- Created Complete Commit API that changes the commitStatus of a 'paid' commit to 'completed'.
- The endpoint is POST /commits/{commitId}/complete.

#### Success Flow
Screenshot 2020-08-06 at 3 58 02 AM
- Also, numCompleted of listing in datastore has increased by 1. 

#### Current `commitStatus` is not 'paid'
Screenshot 2020-08-06 at 4 11 54 AM

Part of #10


server/src/api/commits.ts
```diff
@@ -19,6 +19,7 @@
  */
 
 import express, {Request, Response, NextFunction} from 'express';
+import _ from 'lodash';
```
hzxcaryn, line 22:
```suggestion
import isEqual from 'lodash/isEqual';
```
Prefer this because let's not import the entire library when we only need 1 function! Also, not a fan of `_` because it is a commonly used symbol to refer to an ignored variable (although I know lodash like to refer to themselves as `_` haha). If you like, you can also name it `_isEqual` if you want to be clear the func is frm lodash.
On another hand, is lodash really necessary?

karenfrilya97:
> Could you rename `edit` to `update`? Haha I know I named it that way before the HOF PR but now that that one is merged, I think `update` sounds better! :D

@hzxcaryn Okay!

server/src/api/commits.ts
karenfrilya97: @hzxcaryn As discussed over Chat, will standardise the check with this https://github.com/googleinterns/gpay-group-buy/pull/195/files#diff-5babba68806a5317f63daf9e8dfbd212

ifhuang:
Can you update PR description to add some context about why is this edit commit API needed?

server/src/services/commits.ts
```diff
@@ -125,19 +125,47 @@ const payForCommit = async (
...
+      if (commit.commitStatus !== 'successful') {
+        throw new Error('Only successful commits can be paid.');
+      }
+      break;
+    case 'completed':
+      if (commit.commitStatus !== 'paid') {
+        throw new Error('Only paid commits can be completed.');
+      }
+      break;
+    case undefined:
```
ifhuang, line 162: why do nothing instead of throw error for undefined commit status?

karenfrilya97:
After discussing with @hzxcaryn, decided to change the endpoint to specifically Complete Commit API instead of a more general Edit Commit API. The reason is I thought that RESTful API doesn't allow the use of verbs in the url such as /commits/{commitId}/complete. However, upon looking at commonly used APIs such as [this](https://stripe.com/docs/api/orders/return) we decided that the use of verb on API url can be allowed if it does a specific operation that is not CRUD. Moreover, the specific endpoint to Complete Commit serves our purpose better as it serves our specific purpose, is clearer and prevents clients from making disallowed edit requests.

server/src/api/commits.ts
karenfrilya97: @hzxcaryn We no longer have this check after changing the endpoint to Complete Commit instead of Edit Commit.

karenfrilya97:
> Can you update PR description to add some context about why is this edit commit API needed?

@ifhuang It was to be used to complete a commit. Changed the API to Complete Commit API instead after discussion with Caryn, the explanation is in the comment above.

server/src/services/commits.ts
```diff
@@ -134,12 +134,33 @@ const payForCommit = async (
...
+  return commitStorage.updateCommit(commitId, fieldsToEdit, commit.listingId);
+};
+
+/**
+ * Completes the commit with the specified commitId.
+ * An error will be thrown if the commit is not PAID.
+ * @param commitId Id of the commit to be deleted
+ */
+const completeCommit = async (commitId: number) => {
+  // Check that commit exists.
```
chshda, line 151: What happens if it doesn't exist?

karenfrilya97: @chshda It will just throw an error.

PR #204, opened 2020-08-06
Add missing `await` in addEntityHelper

karenfrilya97:
### Motivation
When adding entity outside of a transaction without the `await`, there are several problems:
- Function returns without resolving the promise/async function. So, any error/promise rejection might not be thrown to the function caller.
- `key` variable does not have `id` property. So, the returned object will be `NaN` instead of the id of the added entity.

On the other hand, adding `await` for transaction causes a deadlock. This is because `transaction.insert()` is resolved only after `transaction.commit()` is resolved. However, `transaction.commit()` cannot be called until `transaction.insert()` is resolved.

To solve this problem, we `await actor.insert()` only if `actor` is `datastore`, not `transaction`.


server/src/storage/datastore.ts
```diff
@@ -268,11 +268,19 @@ const addEntityHelper = async (
...
+        // Do not await here because promise is resolved only after
+        // `await transaction.commit()` is called.
+        actor.insert(entity);
+      }
+    } else {
+      if (actor instanceof Datastore) {
+        await insertUniqueEntity(kind, entity, uniqueProperties);
+      } else if (actor instanceof Transaction) {
+        await uniqueInsertInTransaction(kind, entity, uniqueProperties)(actor);
+      }
```
ifhuang, line 283: now we really have different impl for transaction and datastore, then there is no need to create actor var.

suggest to change to sth. like this:

if (transaction != null) {
  ...
} else {
  ...
}

server/src/storage/datastore.ts
```diff
@@ -268,11 +268,19 @@ const addEntityHelper = async (
 
   try {
     if (uniqueProperties === undefined) {
-      actor.insert(entity);
-    } else if (actor instanceof Datastore) {
-      await insertUniqueEntity(kind, entity, uniqueProperties);
-    } else if (actor instanceof Transaction) {
-      await uniqueInsertInTransaction(kind, entity, uniqueProperties)(actor);
+      if (actor instanceof Datastore) {
```
xlai20, line 271: if (actor instanceof Datastore) {
   if (uniqueProperties != null) ...
   else ...
}

if (actor instanceof Transaction) {
   if (uniqueProperties != null) ...
   else ...
}

server/src/storage/datastore.ts
chshda: +1

karenfrilya97: @ifhuang @chshda Okay! Will do this. I didn't realise that datastore and transaction were separate at first and only combined in the actor var.

PR #205, opened 2020-08-07
Schedule function to update `listingStatus` to 'successful' or 'unsuccessful' when update is reached

### LINKED ISSUE #93: Automatically update listing status when listing deadline is passed

karenfrilya97:
### Changes
- Added a POST endpoint that finds all outdated ongoing listings and update the status to 'successful' or 'unsuccessful'.
- The POST endpoint is invoked by a cron job scheduler (Google Cloud Scheduler) every day at midnight IST.

Closes #93


server/src/services/listings.ts
```diff
@@ -14,16 +14,23 @@
...
+const addListing = async (
+  listing: ListingRequest
+): Promise<ListingResponse> => {
+  const addedListing = await listingStorage.addListing({
     ...listing,
     ...DEFAULT_LISTING_PAYLOAD,
   });
-
+  const {id, deadline} = addedListing;
+  schedule.scheduleJob(deadline, () => updateListingSuccess(id));
```
ifhuang, line 31: how does this work?
will it be persisted and resumed if server stopped and restarted?

server/src/services/listings.ts
```diff
@@ -35,4 +42,17 @@ const getAllListingsWithIds = async (
 const getListing = async (listingId: number): Promise<ListingResponse> =>
   listingStorage.getListing(listingId);
 
+const updateListingSuccess = async (listingId: number) => {
```
ifhuang, line 45: call it updateListingStatus?

server/src/services/listings.ts
```diff
@@ -35,4 +42,17 @@ const getAllListingsWithIds = async (
...
+  const listing = await listingStorage.getListing(listingId);
+  const isSuccessful = listing.numCommits >= listing.minCommits;
+  const affectedCommits = await commitStorage.getAllCommits([
```
ifhuang, line 48: do you need to check status of commit?

server/src/services/listings.ts
xlai20: Same question.

server/src/services/listings.ts
xlai20: Or maybe "updateListingOutcomeStatus" to show that "successful/unsuccessful" status are some kind of terminal state of a listing status.

karenfrilya97: @ifhuang @xlai20 how about `updateListingSuccessStatus`?

server/src/services/listings.ts
karenfrilya97: @ifhuang By right it should all be 'ongoing' at this point. Unless they were manually changed from the Datastore.

server/src/services/listings.ts
karenfrilya97: @ifhuang @xlai20 It won't be persisted. The last time @hzxcaryn and I discussed about this, we thought that we can have a script to reschedule the jobs if the server stopped and restarted.

Alternatively, we can also have cron job, but they will have to run every day, whether or not there is any listing to update and it will also have one additional step of filtering listings which are past deadline and thus need to be updated.

karenfrilya97: But I think we previously didn't consider that GAE might create and shut down instances to manage traffic. In that case, I don't think our solution works. I guess we might still need to have a cron job after all.

server/src/services/listings.ts
xlai20: I think the word "Success" is better to be avoided because your new status can be "unsuccessful" as well. All these are fine:
 - updateListingStatus
 - updateListingTerminalStatus
 - updateListingOutcome
 - updateListingOutcomeStatus

karenfrilya97: @xlai20 I thought 'success' means both 'successful' and 'unsuccessful'. Similar to like if you ask 'What is your level of satisfaction?', then the answer can be 'Satisfied' or 'Unsatisfied'. But if it's confusing, I can change to 'updateListingOutcomeStatus'.

chshda: +1 to avoid the word 'success', it's either unnecessary or confusing to have 'SuccessStatus' compared to just 'Status'.

karenfrilya97: @chshda Okay! I'll change to 'updateListingOutcomeStatus' then.

PR #207, opened 2020-08-07
Add Require Login prompt when login is unsuccessful

hzxcaryn:
# Changes
- Added a prompt that appears after customer refuses to login even after being prompted for actions that requires them to be logged in

Screenshot 2020-08-07 at 9 28 27 PM


web/src/microapps.ts
```diff
@@ -36,8 +36,9 @@ export const decodeToken = (idToken: string) =>
  * Gets microapps identity token and its decoded form.
  * Returns the cached identity if it exists and token has not expired,
  * else, fetches a new identity token from the microapps API.
+ * Returns undefined if calling the getIdentity API fails.
```
ifhuang, line 39: when will this happen? user rejects permission?

web/src/components/customer/contexts/CustomerContext.tsx
```diff
@@ -21,11 +21,16 @@ import useLocalStorage from 'components/common/hooks/useLocalStorage';
...
+  customer: Customer | undefined;
+  idToken: string | undefined;
+};
+
 type ContextType =
   | {
       idToken: string | undefined;
       customer: Customer | undefined;
-      getCustomerWithLogin: () => Promise<Customer | undefined>;
+      getCustomerWithLogin: (x?: boolean) => Promise<AuthenticatedCustomer>;
```
ifhuang, line 33: what is this "x?: boolean"? can you rename x?

hzxcaryn: @ifhuang Oops this `x` is something that I was using for debugging, forgot to remove it from this line! Thanks for pointing this out!

web/src/microapps.ts
hzxcaryn: @ifhuang Yup that's one of the cases that results the getIdentity API to throw an error

PR #209, opened 2020-08-11
Add server http error classes

### LINKED ISSUE #115: Add proper error handling with the appropriate response codes in the api layer
Currently, all thrown errors frm the services/ and storage/ layer have response code 500, which is not correct.
When this issue is being addressed, remember to updated tests error codes as well. Currently they are 
```
expect(res.status).not.toBe(200);
```
change them to test for the exact error codes, such as
```
expect(res.status).toBe(400);
```

hzxcaryn:
Closes #115 

# Changes
- Added `HttpError` base class which comes with the appropriate status code and error message for each error
- Updated tests with the specific error code
  * Also standardised error message phrasing for some error messages 

### TODO in future PRs
- Handle Datastore errors and map each error to a specific `HttpError`.

### DIFF
--- a/server/tests/integration/customers.test.ts
+++ b/server/tests/integration/customers.test.ts
@@ -30,4 +30,27 @@ describe('Customers endpoints', () => {
       expect(res.body).toMatchObject(expectedCustomerData);
     });
   });
+
+  test('Should not fetch by invalid id param', async () => {
+    const customerId = 'invalid-customer-id-type';
+
+    const res = await request(app).get(`/customers/${customerId}`);
+
+    expect(res.status).toBe(400);
+    expect(res.body).toHaveProperty('error');
+    expect(res.body.error.message).toBe(`Invalid customerId ${customerId}`);
+  });
+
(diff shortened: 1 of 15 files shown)


server/tests/integration/customers.test.ts
```diff
@@ -30,4 +30,27 @@ describe('Customers endpoints', () => {
...
+    expect(res.body.error.message).toBe(`Invalid customerId ${customerId}`);
+  });
+
+  test('Should not fetch by non-existent customerId', async () => {
+    const customerId = customerFixtures.ids.length + 1;
+
+    const res = await request(app).get(`/customers/${customerId}`);
+
+    // TODO: Test for actual error status codes when implemented
+    expect(res.status).not.toBe(200);
```
ifhuang, line 50: should assert to be 400 error code?

hzxcaryn: @ifhuang Whether the entity is found or not is thrown by the datastore layer, which is not handled in this PR, so this test is still not changed!

PR #211, opened 2020-08-11
Handle errors in datastore layer

### LINKED ISSUE #210: Handle Datastore errors and throw the appropriate status codes
Currently, datastore errors are caught and thrown as generic errors, with a generic error message (that is not applicable to all possible cases) and status code of 500 (which is not accurate).

Handle datastore errors by their [error codes](https://cloud.google.com/datastore/docs/concepts/errors) and re-throw them as the appropriate `HttpError`

hzxcaryn:
Closes #210 

# Changes
- Handle datastore layer errors and throw them as the appropriate `StorageError`s
- Added `error-mapper` middleware which maps each `StorageError` to a `HttpError` and throws it as the appropriate `HttpError`
- Modified tests to test for the exact status codes

### DIFF
--- a/server/src/storage/datastore.ts
+++ b/server/src/storage/datastore.ts
@@ -19,12 +19,11 @@ import {Entity} from '@google-cloud/datastore/build/src/entity';
 
 import {Filter, OrderRule, StringKeyObject} from '../interfaces';
 import {
-  NotFoundError,
-  InternalServerError,
-  ConflictError,
-  HttpError,
-  BadRequestError,
-} from '../utils/http-errors';
+  StorageError,
+  EntityNotFoundError,
+  EntityAlreadyExistsError,
+  PropertyMismatchError,
+} from '../utils/storage-errors';
 import {UpdateRule} from './interfaces';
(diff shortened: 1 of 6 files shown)


server/src/storage/datastore.ts
```diff
@@ -18,6 +18,13 @@ import {Datastore, Transaction} from '@google-cloud/datastore';
 import {Entity} from '@google-cloud/datastore/build/src/entity';
 
 import {Filter, OrderRule, StringKeyObject} from '../interfaces';
+import {
+  NotFoundError,
+  InternalServerError,
+  ConflictError,
+  HttpError,
```
ifhuang, line 25: ideally, at the storage layer there is no concept of http error. storage error should be thrown and got converted into http error in the upper layer. wdyt?

hzxcaryn: @ifhuang I have the same thoughts too! But when I was implementing it, I'm not sure if I want to add that additional layer of complexity haha :p I'll take another look into this!

PR #212, opened 2020-08-12
Mock auth middlewares and add tests

hzxcaryn:
# Changes
- Mock auth middlewares in integration tests 
  * Test if these middlewares are called for endpoints that require authentication
  * Allow swapping between actual implementation and mock implementation of auth middlewares
- Added implementation tests for customer endpoints to show that our auth mock is working as intended

### DIFF
--- a/server/.eslintrc.json
+++ b/server/.eslintrc.json
@@ -31,7 +31,10 @@
         "**/?(*.)+(spec|test).+(ts|tsx|js)"
       ],
       "rules": {
-        "node/no-unpublished-import": 0
+        "node/no-unpublished-import": 0,
+        // Disable because we this for TS typecasting so that we
+        // can access the helper functions in our mocks.
+        "jest/no-mocks-import": 0
       }
     }
   ]
--- a/server/.eslintrc.json
+++ b/server/.eslintrc.json
@@ -32,8 +32,8 @@
       ],
       "rules": {
         "node/no-unpublished-import": 0,
-        // Disable because we this for TS typecasting so that we
-        // can access the helper functions in our mocks.
+        // Disable because we need this for TS typecasting so that
+        // we can access the helper functions in our mocks.
         "jest/no-mocks-import": 0
       }
     }
--- a/server/tests/integration/commits.test.ts
+++ b/server/tests/integration/commits.test.ts
@@ -36,11 +36,6 @@ beforeAll(() => {
   restoreCustomerAuth();
 });
 
-// Disable customer auth mock implementation by default
-beforeAll(() => {
-  restoreCustomerAuth();
-});
-
 describe('Commits endpoints', () => {
   beforeEach(() => {
     jest.clearAllMocks();
--- a/server/tests/integration/commits.test.ts
+++ b/server/tests/integration/commits.test.ts
@@ -17,13 +17,24 @@
 import request from 'supertest';
 
 import app from '../../src';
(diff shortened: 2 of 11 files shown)


server/tests/integration/commits.test.ts
```diff
@@ -99,4 +105,45 @@ describe('Commits endpoints', () => {
       );
     });
   });
+
+  describe('POST /commits', () => {
+    test('Should require customer auth', async () => {
+      await request(app).post('/commits');
+
+      expect(customerAuth).toHaveBeenCalled();
```
ifhuang, line 113: what is the scope of the mock? will we get a new mock for each test?

if the mock is the same for all test cases, then we need to reset the mock in the setup of each test

server/tests/integration/listings.test.ts
```diff
@@ -127,4 +133,12 @@ describe('Listings endpoints', () => {
       );
```
ifhuang, line 133: can you also add assert here about mock is never called?

server/tests/integration/commits.test.ts
hzxcaryn: @ifhuang Oh right that slipped my mind! Adding it in now! The scope of the mock is for all the tests in the file.

server/tests/integration/listings.test.ts
hzxcaryn: @ifhuang I made a big change to my mock implementation (allow choosing between restoring and mocking auth). By default, actual implementation of the middleware is applied. Hence I think there is no need to assert that auth mock is never called anymore, because if it has been called, then response would have been 401, and any of the tests would not have passed.

server/.eslintrc.json
```diff
@@ -31,7 +31,10 @@
         "**/?(*.)+(spec|test).+(ts|tsx|js)"
       ],
       "rules": {
-        "node/no-unpublished-import": 0
+        "node/no-unpublished-import": 0,
+        // Disable because we this for TS typecasting so that we
```
ifhuang, line 35: nit: Disable because we need this

server/tests/integration/commits.test.ts
```diff
@@ -17,11 +17,35 @@
...
+} = customerAuthMiddleware as typeof mockedCustomerAuth;
+
+// Disable customer auth mock implementation by default
+beforeAll(() => {
+  restoreCustomerAuth();
+});
+
+// Disable customer auth mock implementation by default
+beforeAll(() => {
+  restoreCustomerAuth();
```
ifhuang, line 41: why same restoreCustomerAuth twice?

server/tests/integration/customers.test.ts
```diff
@@ -17,39 +17,183 @@
...
+      });
+
+      test('Should reject requests with invalid phone number', async () => {
+        const customerId = customerFixtures.ids?.[0];
+        const defaultFulfilmentDetails = {
+          name: 'Pusheen Takoyaki',
+          address: 'Disneyland',
+          contactNumber: '+6591234567',
+        };
+        const res = await request(app).patch(`/customers/${customerId}`).send({
```
ifhuang, line 189: add one empty line above

server/tests/integration/commits.test.ts
hzxcaryn: @ifhuang Oops! Must have been left behind from merging and I didn't notice it! Will remove one of them!

PR #213, opened 2020-08-14
Use Spot Messages API to notify customers of `commitStatus` change

### LINKED ISSUE #94: Remind customers to pay for successful listings using Spot's Message API

karenfrilya97:
Closes #94.

### Changes
- Added `sendMessage` handler to call Spot Messages API.
- Added `sendCommitMessage` function to form message and send it using `sendMessage`.
- Called `sendCommitMessage` function from `completeCommit` and `updateListingOutcomeStatus` functions.

### Demo
IMG_20200814_234703_755


server/.env.development
```diff
@@ -4,8 +4,12 @@
...
 SERVER_URL=http://localhost:5000
 WEB_URL=http://localhost:3000
+SPOT_MERCHANT_ID=BCR2DN6T3P5NDAKX
+SPOT_MICROAPP_ID=14116337401048649233
 
 # Secret keys
 # Redefine these in .env.development.local with the appropriate values.
 # Secret environment variables for production should be written in app.secret.yaml.
 OAUTH_CLIENT_ID=<Secret>
+SERVICE_ACCOUNT_CLIENT_EMAIL=<Secret>
```
ifhuang, line 14: do you need to add both service account keys to server/app.yaml?

server/src/interfaces.ts
```diff
@@ -22,6 +22,9 @@ import {Operator} from '@google-cloud/datastore/build/src/query';
  */
 export interface CustomerPayload {
   gpayId: string;
+  // Phone number in the format of country code appended by national number (as defined
+  // by the International Telecommunication Union (ITU)) and separated by space.
```
ifhuang, line 26: can you write an example of such phone number format to the comment?

server/src/utils/spot-api.ts
```diff
@@ -0,0 +1,72 @@
...
+  // by the International Telecommunication Union (ITU)) and separated by space.
+  phoneNumber: {[number: string]: string};
+}
+
+interface Message {
+  imageUrl: string;
+  title: string;
+  text: string;
+  actions: ActionButton[];
+  recipients: PhoneNumberRecipient[];
```
ifhuang, line 41: how will recipients be used?

server/.env.development
karenfrilya97: @ifhuang Yep! We need both there. But since this is a secret, I added it to app.secret.yaml which is not tracked by git.

server/src/utils/spot-api.ts
karenfrilya97: @ifhuang It is used to identify the customer who will receive the message sent using Messages API.

server/src/interfaces.ts
xlai20: +1

server/src/services/commits.ts
```diff
@@ -163,7 +164,63 @@ const completeCommit = async (commitId: number) => {
...
+    title: listing.name,
+    text,
+    actions: [
+      {
+        label: 'View Listing',
+        url: encodeMicroappsUrl(`listing/${listing.id}`),
+      },
+    ],
+    recipients: [
+      {
```
xlai20, line 215: Looks like there is always 1 recipient only for each completeCommitMessage? Can there be a scenario when there are multiple recipients being included in one message?

karenfrilya97: @xlai20 Based on [Messages API docs](https://developers.google.com/pay/spot/eap/reference/messages-api#re-engagement_message), it says that "Only 1 recipient is allowed per request". It's weird that they use an array but, I don't know, maybe it's for future versions where they allow multiple recipients.

server/src/interfaces.ts
hzxcaryn: Actually Karen can we change this to 
```
  // gpayContactNumber format is E.123 international notation
  // Note that the national phone number portion will only contain digits (no spaces)
  // Eg: +91 1234567890
```
It's a description that's partially taken from Spot API docs as well haha. I feel that it's more specific also!

karenfrilya97: @ifhuang @xlai20 Okay!
@hzxcaryn Okay! I put that comment but also added one more line.

PR #214, opened 2020-08-15
Add customer access control middlewares

### LINKED ISSUE #119: Ensure customer is making the request for their own commits when they make POST & DELETE requests on commits
Do that by checking whether the authenticated customer is the same as the customer of the new/existing commit.
An effort to deal with non-happy paths.

hzxcaryn:
Closes #119

# Background
Previously, we only check for authentication of customers, and did not have access control restrictions. This means that as long as I have a valid token, I can make requests to modify another customer's commits and customer data. This is not desirable and this PR fixes this by adding a layer of access control to restricted customer endpoints. 

# Changes
- Added access control middleware that checks if customer is authorized to access/modify the resource they have requested
  * This middleware requires a prior middleware to add an authorized gpay id to `req.authorizedCustomerGPayId` before it can be checked in the `customerAccessControl` middleware. 
(Initially, I wanted `customerAccessControl` middleware to be a higher order function (as seen in the very first commit of this PR), but it was too much of a pain to mock it, and in view of the tight deadline, I decided to change it to this current pattern instead)
- Hoisted validation code in the `api` layer to _above_ the `customerAccessControl` middleware, because we don't want the part that sets `req.authorizedCustomerGPayId` to skip validations
- Added tests & a test mock for the middleware

### DIFF
--- a/server/src/api/customers.ts
+++ b/server/src/api/customers.ts
@@ -63,7 +63,11 @@ customerRouter.post(
     req.validated = req;
     next();
   },
-  customerAccessControl((req: Request) => req.validated.body.gpayId),
+  async (req: Request, res: Response, next: NextFunction) => {
+    req.authorizedCustomerGPayId = req.validated.body.gpayId;
+    next();
+  },
+  customerAccessControl,
   async (req: Request, res: Response, next: NextFunction) => {
     const customerData: CustomerPayload = req.validated.body;
 
@@ -118,12 +122,18 @@ customerRouter.patch(
     };
     next();
   },
-  customerAccessControl(async (req: Request) => {
-    const customer = await customerService.getCustomer(
-      req.validated.params.customerId
-    );
-    return customer.gpayId;
-  }),
+  async (req: Request, res: Response, next: NextFunction) => {
+    try {
+      const customer = await customerService.getCustomer(
+        req.validated.params.customerId
+      );
+      req.authorizedCustomerGPayId = customer.gpayId;
+      next();
+    } catch (err) {
+      next(err);
+    }
+  },
+  customerAccessControl,
   async (req: Request, res: Response, next: NextFunction) => {
     const {customerId} = req.validated.params;
     const fieldsToUpdate = req.validated.body;
--- a/server/src/api/customers.ts
+++ b/server/src/api/customers.ts
@@ -21,6 +21,7 @@
 import {Router, Request, Response, NextFunction} from 'express';
 
 import {CustomerPayload} from '../interfaces';
+import customerAccessControl from '../middleware/customer-access-control';
 import customerAuth from '../middleware/customer-auth';
 import validateAndFormatPhoneNumber from '../middleware/validation/phone-number';
 import {customerService} from '../services';
@@ -58,7 +59,13 @@ customerRouter.post(
   '/',
   customerAuth,
   async (req: Request, res: Response, next: NextFunction) => {
-    const customerData: CustomerPayload = req.body;
+    // TODO: Ensure that req.body is of type CustomerPayload
+    req.validated = req;
+    next();
+  },
+  customerAccessControl((req: Request) => req.validated.body.gpayId),
+  async (req: Request, res: Response, next: NextFunction) => {
+    const customerData: CustomerPayload = req.validated.body;
 
     try {
       // Customers are unique by their gpayId, so we will retrieve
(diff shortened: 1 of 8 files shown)


server/src/middleware/customer-access-control.ts
```diff
@@ -0,0 +1,67 @@
...
+    if (gpayIdOfCurrCustomer === undefined) {
+      throw new UnauthorizedError('User cannot be identified.');
+    }
+
+    const gpayIdOfAuthorizedCustomer = req.authorizedCustomerGPayId;
+    if (gpayIdOfAuthorizedCustomer === undefined) {
+      throw new InternalServerError('req.authorizedCustomerGPayId is not set.');
+    }
+
+    if (gpayIdOfCurrCustomer !== gpayIdOfAuthorizedCustomer) {
```
ifhuang, line 56: so gpayIdOfCurrCustomer is from spot identity api. where is gpayIdOfAuthorizedCustomer from?

server/src/api/customers.ts
```diff
@@ -58,7 +59,17 @@ customerRouter.post(
   '/',
   customerAuth,
   async (req: Request, res: Response, next: NextFunction) => {
-    const customerData: CustomerPayload = req.body;
+    // TODO: Ensure that req.body is of type CustomerPayload
+    req.validated = req;
+    next();
+  },
+  async (req: Request, res: Response, next: NextFunction) => {
+    req.authorizedCustomerGPayId = req.validated.body.gpayId;
```
ifhuang, line 67: looks like req.validated.body.gpayId is also from spot api identity?
if so, then customerAccessControl is not needed here since inside it's comparing two ids from the same source?

server/tests/integration/customers.test.ts
```diff
@@ -89,9 +99,33 @@ describe('Customers endpoints', () => {
...
+
+      const expectedCustomerData = customerFixtures.responseData?.[0];
+      const gpayId = expectedCustomerData.gpayId;
+
+      const res = await request(app).post('/customers').send({gpayId});
+
+      expect(res.status).toBe(403);
+      expect(res.body).toHaveProperty('error');
+      expect(res.body.error.message).toBe(
+        'Not allowed to modify or access requested resource.'
```
ifhuang, line 121: where did you set gpayid from spot identity api/token?

server/src/middleware/customer-access-control.ts
hzxcaryn: @ifhuang 
`gpayIdOfCurrCustomer` is derived from the **decoded Authorization token** that is placed in Header of requests that requires customer auth
`gpayIdOfAuthorizedCustomer` is derived from the request. So for example you want to edit customer 123's data in the request. `gpayIdOfAuthorizedCustomer` will be the gpayId of customer 123.
We will then check if gpayId of customer 123 matches with the gpay id in the authorization token provided by the customer.
Basically this prevents 1 authorized customer from accessing/modifying another authorized customer's data

server/src/api/customers.ts
hzxcaryn: @ifhuang `req.validated.body.gpayId` is the `gpayId` field in the request body. From the client Group Buy app, it is indeed retrieved from api identity, but on the server side, we cannot just trust the input from the request, because another user can just replay the request with their own token, but with my gpayId.

server/tests/integration/customers.test.ts
hzxcaryn: @ifhuang Oh you mean for the test? It is in the customer-auth mock which is in PR #212. Basically I just added a dummy sub for the mock.
image

server/src/middleware/customer-access-control.ts
```diff
@@ -0,0 +1,67 @@
...
+  UnauthorizedError,
+  ForbiddenError,
+  InternalServerError,
+} from '../utils/http-errors';
+
+/**
+ * Authorizes a customer.
+ * Requires req.authorizedCustomerGpayId to be set in order for this
+ * access control middleware to work.
+ * @params getGpayIdOfAuthorizedCustomer Getter function to get gpayId of
```
chshda, line 29: Where is this getGpayIdOfAuthorizedCustomer?

hzxcaryn: @chshda Ah good catch! I removed it because mocks were not cooperating for HOF middlwares, but left out this line when updating!

PR #216, opened 2020-08-17
Update gpayContactNumber upon customer login

hzxcaryn:
Part of #217
To be merged after #215 

# Changes
- Added `gpayContactNumber` field to `CustomerPayload` to allow #213 as the Spot API requires the customer's contact number in gpay in order for our Spot to send messages to the customers
  * We initially wanted to use a [Payment recipient](https://developers.google.com/pay/spot/eap/reference/messages-api#payment-recipient) but we are unable to call that payments API, so we settled with [Phone Number recipient](https://developers.google.com/pay/spot/eap/reference/messages-api#phone-number-recipient)
- When an existing customer signs in with a different `gpayContactNumber` from its previous value, we will update the `gpayContactNumber` of the customer to the newer one 
  * This is because in GPay, customers can change their phone numbers and we will have no idea if the customer changed it or not. Hence, we have decided to send customer's gpay phone number along with the login endpoint, and update it in our database if there is a change.

# Screenshots
The below screenshots show 2 different `gpayContactNumber` body fields when trying to login the same customer (same `gpayId`)
Screenshot 2020-08-17 at 2 36 56 PM
Screenshot 2020-08-17 at 2 37 39 PM

### DIFF
--- a/server/src/api/customers.ts
+++ b/server/src/api/customers.ts
@@ -58,6 +58,12 @@ customerRouter.get(
 customerRouter.post(
   '/',
   customerAuth,
+  validateAndFormatPhoneNumber(
+    (body: Partial<CustomerPayload>) => body.gpayContactNumber,
+    (body: Partial<CustomerPayload>, phoneNumber: string) =>
+      (body.gpayContactNumber = phoneNumber),
+    false
+  ),
   async (req: Request, res: Response, next: NextFunction) => {
     // TODO: Ensure that req.body is of type CustomerPayload
     req.validated = req;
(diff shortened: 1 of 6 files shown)


server/src/api/customers.ts
```diff
@@ -74,10 +80,24 @@ customerRouter.post(
     try {
       // Customers are unique by their gpayId, so we will retrieve
       // the customer using their gpay Id.
-      const existingCustomer = await customerService.getCustomerWithGpayId(
+      let existingCustomer = await customerService.getCustomerWithGpayId(
         customerData.gpayId
       );
       if (existingCustomer !== null) {
+        // Update gpayContactNumber if it does not match
```
xlai20, line 87: Why do you update the contact number when it does not match with existing customer's existing record? Shouldn't you refuse login because of this mismatch?

server/src/api/customers.ts
```diff
@@ -58,6 +58,12 @@ customerRouter.get(
 customerRouter.post(
```
xlai20, line 58: I think you need to update the function comment to include more things that this function is doing.

hzxcaryn: @xlai20 Oh right! Ok sure!

server/src/api/customers.ts
hzxcaryn: @xlai20 User might have changed their phone number on GPay between subsequent logins. We have no way for us to know whether the user has changed their phone number for us to react to that change. The only thing we can do is try to call the getPhoneNumber API to retrieve the current phone number of the customer.

So now the workaround is to get customer's phone number when we request a login and update it on the server side if there is a change in phone number, which is what I am implementing in this PR!

I know that this is not the best because this means users can try to manually create this request to change their gpay phone number and break our messages functionality. But because we rely on Spot API to give us the updated phone number, and this Phone Number API is only available on the client side, I have decided to go with implementing it this way.

hzxcaryn:
@ifhuang 
> what happen if we call spot msg api with a stale customer phone number?

The message is probably will just not get sent :(

PR #222, opened 2020-08-17
Add fallback UI for My Commits page

hzxcaryn:
# Changes
- Add fallback UI for My Commits page
  * If any section is empty, omit the section
  * If there are no commits, show "You have no commits yet" display

Screenshot 2020-08-18 at 12 41 38 AM


web/src/assets/customer/empty.svg
```diff
@@ -0,0 +1,112 @@
+<svg width="192" height="150" viewBox="0 0 192 150" fill="none" xmlns="http://www.w3.org/2000/svg">
```
ifhuang, line 1: if it's only used by commit page then call it empty_commits.svg ?

hzxcaryn: @ifhuang HAHA ok I agree that `empty.svg` sounds bad 😆 I feel that `empty_commits.svg` is too specific because the illustration has nothing to do with commits as well. What about `no_content.svg`? Is that weird too? hahaha

PR #224, opened 2020-08-18
Update listingStatus to 'completed' after completing the last commit

karenfrilya97:
### Changes
- In `updateCommit` function, if the caller wants to change `commitStatus` to `complete` add these additional steps:
  - Check if `affectedListing.numCompleted` will become equal to `numCommits` after completing this commit.
  - If yes, update `affectedListing.listingStatus` to 'completed'.


server/src/storage/commits.ts
```diff
@@ -101,13 +101,22 @@ const updateCommit = async (
...
         op: 'add',
         value: 1,
       });
+
+      const affectedListing = await getEntity(LISTING_KIND, affectedListingId);
+      if (affectedListing.numCompleted + 1 === affectedListing.numCommits) {
+        listingUpdateRules.push({
+          property: 'listingStatus',
+          value: 'completed',
+        });
```
hzxcaryn, line 116: Do we need `op: 'replace'`?
```suggestion
        listingUpdateRules.push({
          property: 'listingStatus',
          op: 'replace',
          value: 'completed',
        });
```

karenfrilya97: @hzxcaryn It's not necessary because the default for `op` is 'replace'. But if it's not clear then I can add the `op` explicitly.

server/src/storage/commits.ts
```diff
@@ -101,13 +101,22 @@ const updateCommit = async (
         value: 1,
       });
       break;
-    case 'completed':
+    case 'completed': {
```
ifhuang, line 104: redundant { ?

ifhuang (approved):

server/src/storage/commits.ts
```diff
@@ -101,13 +101,22 @@ const updateCommit = async (
...
       listingUpdateRules.push({
         property: 'numCompleted',
         op: 'add',
         value: 1,
       });
+
+      const affectedListing = await getEntity(LISTING_KIND, affectedListingId);
+      if (affectedListing.numCompleted + 1 === affectedListing.numCommits) {
```
xlai20, line 112: Add an explanatory comment that after the update of numCompleted add, "numCompleted will be equal to numCommits" which makes the listing status to be completed.

xlai20 (approved):
lgtm with nits

server/src/storage/commits.ts
karenfrilya97: @ifhuang I added the block because there is variable declaration which is block-scoped. Without the block, "the lexical declaration is visible in the entire switch block but it only gets initialized when it is assigned, which will only happen if the case where it is defined is reached." (https://eslint.org/docs/rules/no-case-declarations)

server/src/storage/commits.ts
karenfrilya97: @xlai20 Okay!

PR #233, opened 2020-08-19
Validate listing deadline before adding listing to database

karenfrilya97:
### Changes
- Added `validateAndFormatDate` middleware which checks that listing deadline is a valid date and formats it into a `Date` object.


server/tests/unit/middleware/validation/date.test.ts
```diff
@@ -0,0 +1,138 @@
...
+jest.mock('../../../../src/middleware/merchant-auth');
+
+const {mockMerchantAuth} = merchantAuth as typeof mockedMerchantAuth;
+
+describe('Date validation', () => {
+  beforeAll(() => {
+    mockMerchantAuth();
+  });
+
+  const validListing = {
```
hzxcaryn: Why not take this from listingFixtures?

server/tests/unit/middleware/validation/date.test.ts
```diff
@@ -87,7 +87,9 @@ describe('Date validation', () => {
   });
 
   describe('Listing with no deadline', () => {
-    const {deadline, ...listingWithNoDeadline} = validListing;
+    /* eslint-disable @typescript-eslint/no-unused-vars */
+    const {deadline: _, ...listingWithNoDeadline} = validListing;
+    /* eslint-enable @typescript-eslint/no-unused-vars */
```
hzxcaryn: If only disabling for 1 line can consider using `eslint-disable-next-line`! But for this I think what we can do is just add this as a general rule in `eslintrc.json` like I have done here https://github.com/googleinterns/gpay-group-buy/blob/master/web/.eslintrc.json#L17

karen-celine: @hzxcaryn Ohh! thanks will just disable the rule for _ in general.

server/tests/unit/middleware/validation/date.test.ts
karen-celine: @hzxcaryn Oh ya, I didn't realise I could use that.

server/tests/unit/middleware/validation/date.test.ts
```diff
@@ -0,0 +1,123 @@
...
+  const res = httpMocks.createResponse();
+  const next = sinon.spy();
+
+  describe('Valid listing deadline', () => {
+    const listingWithFormattedDeadline = {
+      ...validListing,
+      deadline: new Date(validListing.deadline),
+    };
+
+    test('Should format deadline into Date object', () => {
```
ifhuang, line 63: usually tests are 3 parts: arrange, act, assert.
pls add an empty line between 2 parts

ifhuang (approved):

